PR #1, opened 2016-02-23
Bluemix object storage fixes

gekola:
* Fix 500 due to port presence in Host header for bluemix object storage
 * Recognize userid as argument

### DIFF
--- a/lib/fog/openstack.rb
+++ b/lib/fog/openstack.rb
@@ -393,7 +393,8 @@ module Fog
                                           :headers => {'Content-Type' => 'application/json'},
                                           :body    => Fog::JSON.encode(request_body),
                                           :method  => 'POST',
-                                          :path    => path
+                                          :path    => path,
+                                          :omit_default_port => true
                                       })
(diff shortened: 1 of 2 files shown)


seanhandley:
Not sure I'm comfortable with this change.

A lot of different systems use the Fog OpenStack implementation and changing things in the headers we send could lead to unpredictable breakages.

Can you describe in some more detail how IBM BlueMix is breaking when used with Fog @gekola?

Ladas:
I've been meaning to tests this, but our keystone v3 integration tests are down due to caching issues, so it might take some time.

But I agree, feels dangerous to change header like this. If should be configurable, so passing e.g. options[:openstack_omit_default_port] and adding it to header only if that config is true

gekola:
The problem with IBM Bluemix Object Storage is when authorizing with `auth_url` being https://lon-identity.open.softlayer.com/v3/auth/tokens (or https://identity.open.softlayer.com/v3/auth/tokens) when Host header contain port (443) the server always returns 500 without further problem explanation.

I agree making the omission optional is a safer way to go. I'll update the PR.

PR #3, opened 2016-02-24
Added some testing jobs and descriptions

jjasghar:
- Added a description for `rake test`
- moved the order of the dependancies
- added rubocop

Now if we decide to use rubocop, running without a `.rubocop.yml` it
comes back with

```
773 files inspected, 12455 offenses detected
```

I suggest we decide on a `.rubocop.yml` before we start doing anything
large with this Gem.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -20,23 +20,6 @@ CollectionMethods:
     reduce: inject
 CyclomaticComplexity:
   Severity: refactor
-GlobalVars:
-  AllowedVariables:
-  # Loggers
-  - $audit_log
-  - $api_log
-  - $aws_log
-  - $fog_log
-  - $log
-  - $miq_ae_logger
-  - $policy_log
-  - $rails_log
-  - $rhevm_log
-  - $kube_log
-  - $scvmm_log
-  - $vim_log
-  # In Automate methods
-  - $evm
 FormatString:
   EnforcedStyle: percent
 HashSyntax:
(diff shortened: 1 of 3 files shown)


Ladas:
the one we use in ManageIQ, https://github.com/ManageIQ/manageiq/blob/master/.rubocop.yml for inspiration :-)

though we take some of them more as recommendations, rewriting everything to 25 methods with low complexity can be hard sometimes :-)

jjasghar:
Just to update, after running that version @Ladas it came back with:

```
773 files inspected, 6592 offenses detected
```

I did have to updated a few of the Cops though:
```
Error: The `Style/TrailingComma` cop no longer exists. Please use `Style/TrailingCommaInLiteral` and/or `Style/TrailingCommaInArguments` instead.
Error: The `Style/SingleSpaceBeforeFirstArg` cop has been renamed to `Style/SpaceBeforeFirstArg.
Error: obsolete parameter RunRailsCops (for AllCops)
```

Ladas:
If nobody is against, I would merge it as a first version, once we remove the ManageIQ specific checks. We can tweak the config later, if it will become more obtrusive then helpful. ;-)

jjasghar:
:+1:

.rubocop.yml
```diff
@@ -0,0 +1,101 @@
...
+StringLiteralsInInterpolation:
+  Enabled: false
+TrailingCommaInLiteral:
+  Enabled: false
+WhileUntilModifier:
+  Enabled: false
+WordArray:
+  AutoCorrect: false
+
+ClassAndModuleCamelCase:
```
Ladas: Seems like you forgot L95-L101

hm I wonder why it doesn't show my comments in the code

Ladas: Ah, I probably commented the commit directly

Ladas:
:+1: cool thank you

PR #6, opened 2016-02-26
Add events to metering

tzumainn:


jjasghar:
Can you add some documentation around this please?

tzumainn:
Sure - in-line, or is there another place for docs I may have missed... ?

jjasghar:
Great question. Probably should start something up. I suggest we create a `docs/` directory off the root, and then maybe start a file called `metering.md`?

We can make md's per each model, which can help out the general understanding of what this gem supports/can do.

tzumainn:
Sounds good!  Is there an existing doc with a format you like, or should I just make one up?

jjasghar:
Yeah go a head and make it up, we can work with whatever you start with and mold it as time progresses.

Something I've hated about fog in general is the lack of documentation, so I'm excited to see that we are making a positive change on our little plugin gem here to help our future selves :metal:.

Ladas:
@tzumainn there is already docs dir with examples https://github.com/fog/fog-openstack/blob/master/lib/fog/openstack/docs/compute.md

@jjasghar it's not wildly used, so maybe just continuing with this and then moving it?

jjasghar:
@ladas oh wow totally missed that. We should probably move that to a better location or reference it in the readme. 

@tzumainn go ahead a use that format for now as we figure out what we want to do. Thanks!

tzumainn:
Ah, cool!  I'll just add to that existing directory and let others figure out if the directory needs to be moved : )

tzumainn:
Added a short doc.  It's far more succinct than the others, primarily because it seemed as if the other docs had a lot of duplicate information.  Let me know if you'd like changes!

jjasghar:
@tzumainn that's PERFECT thank you. I'm good with this merge now. :+1: 

@Ladas any thoughts?

seanhandley:
Nice one @tzumainn :+1:

aufi:
Looks good!

PR #9, opened 2016-02-29
Missing OpenStack requires

Ladas:
Missing requires that were in Fog before
https://github.com/fog/fog/blob/972ae1a990e5ae42a87e3defa85dc3a251818142/lib/fog/openstack.rb

Causing that the classes are not visible

### DIFF
--- a/lib/fog/openstack.rb
+++ b/lib/fog/openstack.rb
@@ -6,6 +6,22 @@ require 'fog/openstack/core'
 require 'fog/openstack/common'
 require 'fog/openstack/errors'
 
+require 'fog/openstack/compute'
+require 'fog/openstack/identity_v2'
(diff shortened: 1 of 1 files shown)


Ladas:
Without these I am getting [NameError]: uninitialized constant Fog::Compute::OpenStack  Method:[rescue in block in refresh] in ManageIQ code

jjasghar:
Makes sense. :+1:

seanhandley:
Interesting. I wonder why the tests passed then?

Ladas:
@seanhandley I think internally, it might not need the classes in tests. Though ManageIQ relies on those classes to be visible.

seanhandley:
Ok @Ladas - let's merge then :+1:

PR #14, opened 2016-03-01
Trigger coveralls after Travis finishes

seanhandley:
Looks like we need this additional step to kick Coveralls off when Travis finishes its test run.


Ladas:
ah, interesting

PR #25, opened 2016-03-02
Bluemix object storage fixes

gekola:
* Fix 500 due to port presence in Host header for bluemix object storage
 * Recognize userid as argument

This is #1 with making port omitting optional.

### DIFF
--- a/lib/fog/openstack.rb
+++ b/lib/fog/openstack.rb
@@ -300,11 +300,12 @@ module Fog
     end
 
     def self.retrieve_tokens_v2(options, connection_options = {})
-      api_key     = options[:openstack_api_key].to_s
-      username    = options[:openstack_username].to_s
-      tenant_name = options[:openstack_tenant].to_s
-      auth_token  = options[:openstack_auth_token] || options[:unscoped_token]
-      uri         = options[:openstack_auth_uri]
+      api_key           = options[:openstack_api_key].to_s
+      username          = options[:openstack_username].to_s
+      tenant_name       = options[:openstack_tenant].to_s
+      auth_token        = options[:openstack_auth_token] || options[:unscoped_token]
+      uri               = options[:openstack_auth_uri]
+      omit_default_port = options[:openstack_auth_omit_default_port]
 
       identity_v2_connection = Fog::Core::Connection.new(uri.to_s, false, connection_options)
       request_body = {:auth => Hash.new}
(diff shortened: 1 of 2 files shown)


seanhandley:
Thanks @gekola. I'm ok with this, though I think you should probably pass that option through to the Keystone v2 method as well.

gekola:
Thanks for the feedback, @seanhandley. I updated the PR to also apply port omitting option to the Keystone v2 method.

seanhandley:
Ok great. And what's the motivation for putting the userid attribute into the storage class?

seanhandley:
I ask because I'm not sure why it isn't present in the compute, image, network etc

gekola:
@seanhandley while working with Carrierwave and Paperclip there is a warning `[fog][WARNING] Unrecognized arguments: openstack_userid` on each file uploading/downloading. Adding userid to parameters recognized by storage removes this warning (I figured since domain and username are there it is OK to put userid there also). I am new to fog, so I am not sure why it is not there for compute, image and so on. But I guess it was not added to them when Keystone v3 support was added while it should have been.

seanhandley:
Ok thanks @gekola - Thumbs up from me :+1: 

We'll maybe look at re-architecting the gem a little so the parameters aren't duplicated between OpenStack projects.

lib/fog/openstack.rb
```diff
@@ -289,6 +289,7 @@ def self.retrieve_tokens_v2(options, connection_options = {})
       tenant_name = options[:openstack_tenant].to_s
       auth_token  = options[:openstack_auth_token] || options[:unscoped_token]
       uri         = options[:openstack_auth_uri]
+      omit_default_port = options[:openstack_auth_omit_default_port]
```
Ladas: indentation nit, we should have rubocop soon, could you indent all = the same? Applies to all below

@seanhandley hm, what are we missing to get rubocop running for each PR?

lib/fog/openstack.rb
```diff
@@ -310,7 +311,8 @@ def self.retrieve_tokens_v2(options, connection_options = {})
         :headers  => {'Content-Type' => 'application/json'},
         :body     => Fog::JSON.encode(request_body),
         :method   => 'POST',
-        :path     => (uri.path and not uri.path.empty?) ? uri.path : 'v2.0'
+        :path     => (uri.path and not uri.path.empty?) ? uri.path : 'v2.0',
+        :omit_default_port => omit_default_port
```
Ladas: I think it might be better to not even list the :omit_default_port if it's nil, so we don't change the current behavior

Could you store the request into {}

then 

request[:omit_default_port] = omit_default_port unless omit_default_port.nil? # that should put here true or false
response = identity_v2_connection.request(request)

same for the keystone v3

Ladas:
@gekola just last comments, then I am ready to merge :-)

seanhandley:
@Ladas You need to run rubocop from Travis on each build afaik but I've never done it myself (I use a custom integration with RubyCritic and Slack for my commercial projects).

From the looks, you can reference rubocop in the .travis.yml and then pass it arguments to configure how strict you need it to be.

gekola:
@seanhandley @Ladas you can see an example of using rubocop with travis in [rubocop repository](https://github.com/bbatsov/rubocop/blob/master/.travis.yml#L28). I do not think adding rubocop check to CI it is a good idea at this point as there are too many offenses in the code, so the build would always fail.

Ladas:
@gekola awesome, thank you :-)

@gekola @seanhandley btw. for ManageIQ the rubocop runs only on the lines changed by the PR, so we can incrementally make it better. I will ask the guys how to setup that.

PR #27, opened 2016-03-04
add volume type and metadata options in create_volume

zhitongLBN:
according to http://developer.openstack.org/api-ref-compute-v2.1.html   --> /v2.1/​{tenant_id}​/os-volumes
added `volume_type` and `metadata` in `compute/create_volume` option

### DIFF
--- a/lib/fog/openstack/requests/compute/create_volume.rb
+++ b/lib/fog/openstack/requests/compute/create_volume.rb
@@ -24,10 +24,10 @@ module Fog
           end
 
           request(
-            body: Fog::JSON.encode(data),
-            expects: [200, 202],
-            method: 'POST',
-            path: 'os-volumes'
+            :body    => Fog::JSON.encode(data),
+            :expects => [200, 202],
+            :method  => 'POST',
+            :path    => 'os-volumes'
           )
         end
       end
--- a/lib/fog/openstack/requests/compute/create_volume.rb
+++ b/lib/fog/openstack/requests/compute/create_volume.rb
@@ -37,21 +37,19 @@ module Fog
         def create_volume(name, description, size, options = {})
           response = Excon::Response.new
           response.status = 202
-          data = {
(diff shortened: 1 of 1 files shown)


lib/fog/openstack/requests/compute/create_volume.rb
```diff
@@ -11,21 +13,28 @@ def create_volume(name, description, size, options={})
...
+
+          vanilla_options.select { |o| options[o] }.each do |key|
             data['volume'][key] = options[key]
           end
+
           request(
-            :body     => Fog::JSON.encode(data),
-            :expects  => [200, 202],
-            :method   => 'POST',
-            :path     => "os-volumes"
```
Ladas: please keep the hash rocket notation =>

zhitongLBN:
done :)

zhitongLBN:
please configure rubocop syntax preference :)

Ladas:
@zhitongLBN A second, I just activated it, sorry for the spam

For some reason it takes the default from houndci, instead of our .rubocop.yml

Ladas:
@zhitongLBN the hound is not listening :-) please ignore the rubocop change for now

Ladas:
@zhitongLBN alright, seems like it started to work, but you might need to change something and push -f, so it restarts the hound job

Ladas:
@zhitongLBN sorry for the complications :-)

zhitongLBN:
@Ladas I did a rebase and push -f

Ladas:
@zhitongLBN those seems like real rubocop issues according to the rules we have set, but the double quote is not required now so you can delete that commit and you should get less conflicts

Btw. if you will think some of the rules are more obtrusive than helpful, start an issue for the change of the rules and we can discuss that :-)

zhitongLBN:
@Ladas adapted to new rubocop style

Ladas:
@zhitongLBN looks good, thank you :-)

Issue #56, opened 2016-03-10
Fog OpenStack keystone v3 caching is non deterministic


plribeiro3000:
brought by:
https://github.com/fog/fog/pull/3738

My biggest problem is, that the behavior is non deterministic now, which is not good for lib I think. So for ManageIQ CI, which is doing a lot of request per one recorded VCR, it will always pass the 30 second interval, always having variable amount of requests cached. Then of course the recorded VCR is much quicker, so that is another issue, it would require having sleep there.

So, this makes it impossible to get deterministic behavior. When Invoking the same methods, I am getting basically random API requests fired. And that always breaks our CI.

My proposed solution:
1.) Having it configurable, disabled caching by default, we would pass e.g. :cache_ttl => 30 as options for auth request, then other services could take the settings from there, e.g. for keystone project list, etc.
So it would be Fog::Compute.new({..., :cache_ttl => 30 })

2.) Revert the caching here and rather rely on application specific implementation of caching, using redis, memcache, etc. But unless we will just have http layer caching, we would need to expose the Auth object.

So e.g. something like this:
auth = Fog::Auth.new({
    :provider            => 'openstack',                                      # OpenStack Fog provider
    :openstack_auth_url  => 'http://KEYSTONE_HOST:KEYSTONE_PORT/v2.0/tokens', # OpenStack Keystone endpoint
    :openstack_username  => OPEN_STACK_USER,                                  # Your OpenStack Username
    :openstack_tenant    => OPEN_STACK_TENANT,                                # Your tenant id
    :openstack_api_key   => OPEN_STACK_PASSWORD,                              # Your OpenStack Password
    :connection_options  => {}                                                # Optional
})

then:

Fog::Compute.new(auth_object => auth)
Fog::Image.new(auth_object => auth)

I think these would not need to do API request then, if we will fetch auth token and keystone catalog into the auth object. And the Auth object could support rescoping maybe? Which we use for changing a tenant scope e.g. Fog::Image.new(auth_object => auth, openstack_tenant => 'project1'), which would just fire one api request for rescoping using existing token.

Original issue opened by @Ladas at fog/fog#3812

plribeiro3000:
cc/ @Ladas @dhague @geemus

kjorg50:
I am still facing this issue when reading OpenStack domains. Subsequent calls to read domains return different results, for example I run this code

```
identity = Fog::Identity::OpenStack.new(@connection_params)

puts "First listing of domains: #{identity.domains}"
[...]

PR #66, opened 2016-03-19
Fix logic to detect identity version

mtakayuki:
Hello.

If we specify an identity v3 URL as `:openstack_auth_url` parameter in `~/.fog` file, `Fog::Identity::Openstack.new` method returns an instance of identity v2, not v3. It occurs because the configurations has not been read from the file, when `Fog::Identity::Openstack.new` method is invoked. So, `is_v3? args` in the method becomes `false` regardless of URL.

In this modification, identity version is detected after configurations are read from `~/.fog` file.

### DIFF
--- a/lib/fog/openstack/identity.rb
+++ b/lib/fog/openstack/identity.rb
@@ -3,25 +3,119 @@
 module Fog
   module Identity
     class OpenStack < Fog::Service
+      requires :openstack_auth_url
+      recognizes :openstack_auth_token, :openstack_management_url, :persistent,
+                 :openstack_service_type, :openstack_service_name, :openstack_tenant,
+                 :openstack_endpoint_type, :openstack_region, :openstack_domain_id,
+                 :openstack_project_name, :openstack_domain_name,
+                 :openstack_user_domain, :openstack_project_domain,
+                 :openstack_user_domain_id, :openstack_project_domain_id,
+                 :openstack_api_key, :openstack_current_user_id, :openstack_userid, :openstack_username,
+                 :current_user, :current_user_id, :current_tenant,
+                 :provider, :openstack_identity_prefix, :openstack_endpoint_path_matches
 
       # Fog::Identity::OpenStack.new() will return a Fog::Identity::OpenStack::V2 or a Fog::Identity::OpenStack::V3,
       #  depending on whether the auth URL is for an OpenStack Identity V2 or V3 API endpoint
       def self.new(args = {})
-        @openstack_auth_uri = URI.parse(args[:openstack_auth_url]) if args[:openstack_auth_url]
         if self.inspect == 'Fog::Identity::OpenStack'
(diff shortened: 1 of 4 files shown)


seanhandley:
Thanks for the PR @mtakayuki.

We need to rewrite the logic for auth so it doesn't rely on regexing the auth URL.

In our production OpenStack environment we're using both Keystone v2 and v3. This is a valid configuration and the endpoint auth URL is simply of the form `http://openstack:35357/`. OpenStack operators can set the endpoint URLs to anything so it's naiive to assume we can rely on the version information being in the URL.

As a stop-gap, I added the `:openstack_identity_prefix` config option, which you can set to `v2.0` or `v3` or whatever happens to be the relevant version in the URL.

It's not an ideal solution, but it works for the time being. I think the ideal method to handle this is to read the API version from the service catalog but I haven't had time to get into it properly yet.

mtakayuki:
Thanks for your comment, @seanhandley.

OK, I try to rewrite the logic.

I didn't recognize the `:openstack_identity_prefix` option. And I agree with your ideal method. After I think to find out the best way to do it, I'll fix the logic.

mtakayuki:
Hello @seanhandley,

I changed the logic.

Basically an identity version is decided from URL path in service catalog, unless `Fog::Identity::OpenStack::V2/3` class is directly specified to instantiate an identity service. When an identity service instance is created, an instance of `Fog::Identity::OpenStack::Real` is created at first. The initialization logic is almost same as one of previous `Fog::Identity::OpenStack::V2/3::Real` except for default service type. After that, `Fog::Identity::OpenStack::V2::Real` or `Fog::Identity::OpenStack::V3::Real` instance is created according to `@path` and its instance variables are just copied from the `Fog::Identity::OpenStack::Real` instance.

seanhandley:
Thanks @mtakayuki - this looks good. We'll review soon.

mtakayuki:
I just fixed up and rebased.

seanhandley:
Please would you address the styling issues that Hound CI has raised and rebase again?

lib/fog/openstack/identity.rb
```diff
@@ -3,25 +3,115 @@
...
+
+        def config_service?
+          true
+        end
+
+        def config
+          self
+        end
+
+        private
```
mtakayuki: I had already added a blank line after `private`.

mtakayuki:
I had already added a blank line after `private`, but the issue does not disappear. I think the reason why I did not modify the line itself and only added a blank line after it. I confirmed my local rubocop did not raise the issue.

By the way, I ignore another issue, i.e. `Use %w or %W for an array of words` because other tests have same issue. Should I fix this issue?

Ladas:
Let's ignore the %w for now. @seanhandley do we want to keep this rule?

Looks good to me, I can't run this through my keystone v3 tests, until I fix this https://github.com/fog/fog-openstack/issues/56. I will just verify it later then, lets move forward for now. :-)

Thank you for the contribution. :-)

seanhandley:
:+1:

PR #63, opened 2016-03-18
added associate and disassociate into floating_ip objet

zhitongLBN:
it was merge in PR https://github.com/fog/fog/pull/3850, but it was during openstack separated from fog,
so please merge this to keep it update, thx:)

PR #67, opened 2016-03-21
Rework floating ip

zhitongLBN:
new PR from #63
* added associate and disassociate into floating_ip objet


seanhandley:
Thanks @zhitongLBN - looks good :+1:

PR #71, opened 2016-04-04
move code to raise service not found

zhitongLBN:
following discussion on https://gitter.im/fog/fog-openstack
move code to check service existent before it break from getting something from `nil`


Ladas:
looks good :+1:

PR #72, opened 2016-04-04
fix mock volume status when status = 'creating'

KevinLoiseau:
With this modification we can use `wait_for { ready? }` after creating volume when we use mocked data.


Ladas:
sounds sane :+1:

PR #76, opened 2016-04-06
Fix Collection when body is empty

gildub:
Adds a case to deal when the response body is empty.
For instance with rules_collection (with coming introspection models) a request
makes possible to delete all rules and it returns `body => ""`


seanhandley:
That sounds like an API inconsistency. I think we should present it to the user as `[]` from the models ultimately.

gildub:
That's what the fix does, allowing to pass when object is empty.
```ruby
nspector.rules_collection.destroy_all
=> []
```

seanhandley:
Ok, I see. Feels like this should be reflected in our tests.

gildub:
Ok, removing the call to load_response to avoid the issue. 
Therefore closing this one.

PR #79, opened 2016-04-14
Add introspection models

gildub:
Models and Collection for Introspection Rules

### DIFF
--- a/lib/fog/openstack/models/introspection/rules_collection.rb
+++ b/lib/fog/openstack/models/introspection/rules_collection.rb
@@ -1,0 +1,32 @@
+require 'fog/openstack/models/collection'
+require 'fog/openstack/models/introspection/rules'
+
+module Fog
+  module Introspection
+    class OpenStack
+      class RulesCollection < Fog::OpenStack::Collection
+        model Fog::Introspection::OpenStack::Rules
+
+        def all(_options = {})
+          load_response(service.list_rules, 'rules')
+        end
+
+        def get(uuid)
+          data = service.get_rules(uuid).body
+          new(data)
+        rescue Fog::Introspection::OpenStack::NotFound
+          nil
+        end
+
+        def destroy(uuid)
+          rules = get(uuid)
+          rules.destroy
+        end
+
+        def destroy_all
+          service.delete_rules_all
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


seanhandley:
👍

lib/fog/openstack/models/introspection/rules_collection.rb
```diff
@@ -0,0 +1,36 @@
...
+    class OpenStack
+      class RulesCollection < Fog::OpenStack::Collection
+        model Fog::Introspection::OpenStack::Rules
+
+        def all(_options = {})
+          load_response(service.list_rules, 'rules')
+        end
+
+        def summary(_options = {})
+          load_response(service.list_rules, 'rules')
```
Ladas: you can omit this, when it's the same as all. Summary delegates to all in the superclass

lib/fog/openstack/models/introspection/rules_collection.rb
```diff
@@ -0,0 +1,36 @@
...
+        end
+
+        def get(uuid)
+          data = service.get_rules(uuid).body
+          new(data)
+        rescue Fog::Introspection::OpenStack::NotFound
+          nil
+        end
+
+        def destroy(uuid)
```
Ladas: just a note: we might just define this method on the superclass, seems like it's generic enough

Ladas:
@gildub just a minor nit, otherwise LGTM :+1:

lib/fog/openstack/models/introspection/rules_collection.rb
gildub: I missed that, thanks!

lib/fog/openstack/models/introspection/rules_collection.rb
gildub: I understand the idea but I'm not following you 'cause the superclass only has method place holders. Are you willing to change that? Nevertheless if we change the superclass, I'd rather do that on a separate patch.

Ladas: sure, this was a side note, so separate patch is ok. Yeah the superclass could hold a common functionality. Less code in the specific classes will be better. :-)

Ladas:
looks great now :+1:

PR #81, opened 2016-04-19
Do not send blank attributes through API

Ladas:
In Liberty API, sending nil values is not supported anymore.
Fixes for this are being done, but not everywhere. So we should
not send any parameter that was not specified.


gildub:
Looks good to me!

PR #85, opened 2016-04-22
Make caching for keystone v3 optional and configurable

Ladas:
Make caching for keystone v3 optional and configurable

Fixes GH issue:
https://github.com/fog/fog-openstack/issues/56

### DIFF
--- a/lib/fog/openstack.rb
+++ b/lib/fog/openstack.rb
@@ -365,6 +365,7 @@ module Fog
       auth_token        = options[:openstack_auth_token] || options[:unscoped_token]
       uri               = options[:openstack_auth_uri]
       omit_default_port = options[:openstack_auth_omit_default_port]
+      cache_ttl         = options[:openstack_cache_ttl] || 0
 
       connection = Fog::Core::Connection.new(uri.to_s, false, connection_options)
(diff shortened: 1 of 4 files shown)


lib/fog/openstack.rb
```diff
@@ -433,7 +434,7 @@ def self.retrieve_tokens_v3(options, connection_options = {})
         request[:omit_default_port] = omit_default_port unless omit_default_port.nil?
 
         response = connection.request(request)
-        @@token_cache[{body: request_body, path: path}] = response, Time.now + 30 # 30-second TTL, enough for most requests
+        @@token_cache[{:body => request_body, :path => path}] = response, Time.now + 30 if cache_ttl > 0
```
gildub: Maybe I missed something, but why not
`@@token_cache[{:body => request_body, :path => path}] = response, Time.now + cache_ttl if cache_ttl > 0`

gildub:
Looks good, just one comment/question above.
As we discussed another approach to the overall problem will be a long term solution but that's for another PR.

lib/fog/openstack.rb
Ladas: ah, yeah, missed that one

PR #87, opened 2016-04-28
changed path to use servers.json for block_device_mapping_v2

dkoper:
block_device_mapping_v2 comes under API POST /servers:
http://developer.openstack.org/api-ref-compute-v2.1.html#createServer

This was failing on our Icehouse based installation where the undocumented /os-volumes_boot API is not supported.

### DIFF
--- a/lib/fog/openstack/requests/compute/create_server.rb
+++ b/lib/fog/openstack/requests/compute/create_server.rb
@@ -77,7 +77,7 @@ module Fog
             end
           end
 
-          path = block_device_mapping ? 'os-volumes_boot.json' : 'servers.json'
+          path = options['block_device_mapping'] ? 'os-volumes_boot.json' : 'servers.json'
 
           request(
             :body     => Fog::JSON.encode(data),


Ladas:
hm, weird, the current code is passing in our test from Icehouse  to Liberty

dkoper:
With the change as well, right?

Ladas:
I will have to test, AFAIK this code was working for us from Grizzly. What was the failure you were seeing exactly?

Ladas:
seems like we are still doing the same thing as python-novaclient, so it should be valid https://github.com/openstack/python-novaclient/blob/3bed5446d7b15901ba8956e5331bcc2054fdd3a9/novaclient/tests/unit/v2/test_servers.py#L194

And seems like python-openstackclient doesn't support v2 mapping at all, so not sure if they are changing it.

I think we were duplicating what nova client is doing, rather than following the docs. Which is usually smart thing to do, given docs can be misleading from times to times. :-)

Ladas:
hm, interesting, seems like os-volumes_boot is just an alias to servers, https://github.com/openstack/nova/blob/b256cae8e204fbbf6f3d40f5f4d47013be018a6d/nova/api/openstack/compute/legacy_v2/contrib/volumes.py#L605

voelzmo:
@Ladas @dhague @dkoper Is this ready to be merged in or do you need some additional clarifications? We'd like to get that fix into the BOSH OpenStack CPI, so we don't use deprecated APIs more than necessary :)

Ladas:
alright, given os-volumes_boot is just an alias to the servers, we should be save to merge this in.

I will run additional tests on our old OpenStack envs later, but I don't expect any problems.

Thank you for the contribution. :+1:

PR #88, opened 2016-04-28
Adds services#get methods with some services tests

petrblaho:
OpenStack API currently does not support getting single service by its id.

This feature is wanted in https://github.com/ManageIQ/manageiq/pull/7996 so it seams better to have this ```find``` block here in fog-openstack with a chance to upgrade it to proper request when (and if) OpenStack API supports it.

### DIFF
--- a/lib/fog/openstack/models/compute/services.rb
+++ b/lib/fog/openstack/models/compute/services.rb
@@ -17,6 +17,18 @@ module Fog
           Fog::Logger.deprecation('Calling OpenStack[:compute].services.details is deprecated, use .services.all')
           all(options)
         end
+
+        def get(service_id)
+          # OpenStack API currently does not support getting single service from it
+          # There is a blueprint https://blueprints.launchpad.net/nova/+spec/get-service-by-id
+          # with spec proposal patch https://review.openstack.org/#/c/172412/ but this is abandoned.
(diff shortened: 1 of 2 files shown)


lib/fog/openstack/models/compute/services.rb
```diff
@@ -17,6 +17,20 @@ def details(options = {})
...
+        def get(service_id)
+          # OpenStack API currently does not support getting single service from it
+          # There is a blueprint https://blueprints.launchpad.net/nova/+spec/get-service-by-id
+          # with spec proposal patch https://review.openstack.org/#/c/172412/ but this is abandoned.
+          serv = service.list_services.body['services'].detect do |s|
+            s['id'] == service_id
+          end
+          if serv
+            new(serv)
+          end
```
gildub: According to rubocop the following would be just fine: `new(serv) if serv`

gildub:
Looks good, besides a hound and a small cosmetic issues.

lib/fog/openstack/models/compute/services.rb
petrblaho: Done.

petrblaho:
@gildub thank you for a review.

I updated branch with modified code.

Hound comments were addressed in the earlier iteration :-)

PR #89, opened 2016-05-03
A couple fixes for evacuate_server

sseago:
1) don't pass in params with nil values
2) use adminPass instead of admin_password, as that's what the nova API expects

### DIFF
--- a/lib/fog/openstack/requests/compute/evacuate_server.rb
+++ b/lib/fog/openstack/requests/compute/evacuate_server.rb
@@ -2,13 +2,13 @@ module Fog
   module Compute
     class OpenStack
       class Real
-        def evacuate_server(server_id, host, on_shared_storage, admin_password = nil)
+        def evacuate_server(server_id, host = nil, on_shared_storage = nil, admin_password = nil)
+          evacuate = {}
+          evacuate['host'] = host if host
+          evacuate['onSharedStorage'] = on_shared_storage if on_shared_storage
+          evacuate['adminPass'] = admin_password if admin_password
           body = {
-            'evacuate' => {
-              'host'            => host,
-              'onSharedStorage' => on_shared_storage,
-              'admin_password'  => admin_password,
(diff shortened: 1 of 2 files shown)


lib/fog/openstack/requests/compute/evacuate_server.rb
```diff
@@ -3,12 +3,12 @@ module Compute
     class OpenStack
       class Real
         def evacuate_server(server_id, host, on_shared_storage, admin_password = nil)
+          evacuate_hash = {}
```
gildub: Why not simply use `evacuate` as the variable name?

seanhandley: Agreed. Type information in variable names goes against the Ruby style guide.

seanhandley: Also, if `host` and `on_shared_storage` are optional then set them to nil in the method signature, then the method can simply be called as `evacuate_server(id)` instead of `evacuate_server(id, nil, nil)`.

sseago: Works for me on the variable name change (I've seen this style used often in the past, but yeah, I'll fix it). I can change the signature too. I was trying to minimize external changes to what was essentially a bug fix here, but if there are no objections to making those all default to nil, that makes more sense to me anyway.

Ladas: also if the parameters are optional, change also signature of https://github.com/sseago/fog-openstack/blob/evacuate_server_api_fix/lib/fog/openstack/models/compute/server.rb#L305

Ladas:
lgtm :+1:

sseago:
I'm going to have to file a follow-on PR. Upon further testing, when onSharedStorage is false, the API requires an explicit false value passed in rather than omitting the param when it's false.

PR #90, opened 2016-05-03
A collection of patches & enhancements from SAP, plus VCR tests enabled

dhague:


Ladas:
huuge, this will take a while to review :-)

dhague:
sorry about that - I've been meaning to integrate these for ages, but there was always something more important in the to-do list

gildub:
Thank you for the effort, this is great although effectively quite big.
Meanwhile I think there are few issues.

- There are 22 commits, I believe the standard practice is to use only one otherwise the logs would be messy after the merge. Could you consolidate?

- Could this be divided in several pull requests in order to make review/validation easier for everyone?

- Also this introduces Rspec and according to fog/fog-core we are going with Minitests:
Per fog/fog#1266 and fog/fog#2630, Minitest is the new testing framework to be used by fog projects.

dhague:
@gildub You're right, 22 changes is a bit much - I will have a go at consolidating them into a more reasonable set of commits with a PR for each.
On the RSpec issue, I am not actually introducing RSpec here - in fact RSpec tests have been in place since the original OpenStack Identity v3 contributions to Fog, and were also included in the initial import to fog-openstack - it's just that in fog-openstack they were not run from the Rakefile for some reason, and I have fixed that.

dhague:
Closing this pull request - individual PRs for each OpenStack service to follow.

PR #91, opened 2016-05-03
Update Port doesn't allow you to modify allowed_address_pairs

rb1whitney:
The fog-openstack driver doesn't allow allowed_address_pairs to be modified once they are set (create_port allows address_pairs). This behavior is desirable once you have a high availability pairing and need to update if your servers/address pair change. The Openstack API allows this call and works perfectly. I have been using this change for a while in a local project and would like to see this feature officially pulled in.

If update_port isn't updated, fog openstack thinks updating allowed address pairs is invalid and fails code such as:
network_driver.update_port(port['id'], {:allowed_address_pairs => [{:ip_address => server['ipaddress']}, {:ip_address => vip}]})

### DIFF
--- a/lib/fog/openstack/requests/network/update_port.rb
+++ b/lib/fog/openstack/requests/network/update_port.rb
@@ -6,7 +6,7 @@ module Fog
           data = { 'port' => {} }
 
           vanilla_options = [:name, :fixed_ips, :admin_state_up, :device_owner,
-                             :device_id, :security_groups]
+                             :device_id, :security_groups, :allowed_address_pairs]
           vanilla_options.select{ |o| options.key?(o) }.each do |key|
             data['port'][key] = options[key]
           end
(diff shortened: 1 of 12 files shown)


gildub:
Thanks for that.

Could you please fix the hound and align the assignment?

Also, any chance you could add some documentation (/lib/fog/openstack/docs) and/or examples (/lib/fog/openstack/examples)? I know there is much for the network part but that's no excuse ;-)

lib/fog/openstack/requests/network/update_port.rb
```diff
@@ -30,6 +30,7 @@ def update_port(port_id, options = {})
             port['device_owner']      = options[:device_owner]
             port['device_id']         = options[:device_id]
             port['security_groups']   = options[:security_groups] || []
+            port['allowed_address_pairs']   = options[:allowed_address_pairs] || []
```
Ladas: please align the = with the others, also you are modifying only tests

rb1whitney:
I just committed an example for networking. The example code creates a new port and adds it to an existing device. I also added the main change that was missing on my initial check-in. Formatting issues should be resolved.

gildub:
Excellent, thank you.
Before we merge, could you please consolidate a bit your commits?
Thanks

rb1whitney:
I just squashed changes into a single commit. Can you use the latest now?

PR #92, opened 2016-05-04
Fixing bug in prior evacuate fix for onSharedStorage

sseago:
Further testing indicated a problem with the recent fixes to evacuate.
Contrary to the OpenStack CLI (where omitting --on-shared-storage is
a 'false' setting), the API treats this as a required param, needing
an explicit true or false.

This PR fixes that, defaulting the param to true


Ladas:
yeah, I guess this will work, the interface could use some clean up later though, :+1:

PR #82, opened 2016-04-19
Restoring spec tests with minitest/spec

gildub:
This part of an effort to consolidate unit tests and to use minitest per
fog-core project.

/spec is containing lot of tests using vcr/webmock using Rspec. This has now
been replaced Rspec with minitest. `minitest_helper.rb` file has been merged
with `spec_helper.rb`.

In the meantime all tests are merged under same umbrella. The /spec unit tests
can be executed using the added Rake task `spec`.

The `vcr` gem's cassettes path seemed inconsistent. There are now located in
/spec/fixtures/openstack.

The `turn` gem has been removed. Used for test results comestic, it's obsolete
and not maintained anymore.

All the existing tests are now passing (but 2, see below), restoring 66 tests,
409 assertions, 0 failures, 0 errors, 11 skips.

Note: There are 2 `identity_v3_spec.rb` which are on errors (not sure they were
green or not). They have been marked as skipped and commented.

PR #93, opened 2016-05-04
Identity API fixes

dhague:
version path fix
prefer project ID over project name
fix caching typo
update VCR recordings & default values

### DIFF
--- a/lib/fog/openstack.rb
+++ b/lib/fog/openstack.rb
@@ -220,13 +220,16 @@ module Fog
               :method => 'GET'
           }
           user_id = body['token']['user']['id']
-          response = Fog::Core::Connection.new(
-              "#{uri.scheme}://#{uri.host}:#{uri.port}/v3/users/#{user_id}/projects", false, connection_options).request(request_body)
+          project_uri = uri.clone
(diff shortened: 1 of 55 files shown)


gildub:
Look good

Ok, re waiting upon #82 and rebase after it has been merged.

One more hound to fix: replacing simple quotes by double quotes. I use to do the same but single ones seems out of trend now, well according to Rubocop!

dhague:
The single quotes thing is annoying - RubyMine suggests using single quotes where there is no interpolation, which seems to be better style to me. I will fix and push an update rebased on #82 once it's merged.

gildub:
Yeah I agree with RubyMine, that's what I been doing to, Grrr!
Thanks

seanhandley:
It used to be that strings with double quotes were slower because the interpreter needed to check for interpolation. As far as I know that's no longer the case so it really doesn't matter from a technical perspective.

Since it makes no real difference, I can see the argument for double quotes all the way through the codebase but frankly it's bikeshedding at best and really doesn't matter 😄

gildub:
@seanhandley, thanks for sharing this interesting little story.

PR #95, opened 2016-05-04
Compute API fixes & enhancements

dhague:
Add volume attachments and list availability zones
Implement update and delete os-extra-specs for flavors
Until networking is complete the API returns a nil value for addresses
Rename Compute "create_volume_snapshot" to "create_snapshot" to avoid confusion with Volume API snapshots
Remove shindo test for non-existent mock


lib/fog/openstack/models/compute/services.rb
```diff
@@ -17,18 +17,6 @@ def details(options = {})
...
-          # OpenStack API currently does not support getting single service from it
-          # There is a blueprint https://blueprints.launchpad.net/nova/+spec/get-service-by-id
-          # with spec proposal patch https://review.openstack.org/#/c/172412/ but this is abandoned.
-          serv = service.list_services.body['services'].detect do |s|
-            s['id'] == service_id
-          end
-          new(serv) if serv
-        rescue Fog::Compute::OpenStack::NotFound
-          nil
-        end
```
gildub: Why removing this one, besides the fact it was just added recently?

gildub:
Many goods things although I believe having a pull request per feature change/add/remove is better, Even though the amount of files impacted is not huge, I would say to break down that one again.

Why is get removed from lib/fog/openstack/models/compute/services.rb?

Could you please address hounds issues?

dhague:
Closing this PR - individual PRs to follow

PR #97, opened 2016-05-04
Network API fixes

dhague:
Add router_id attribute
Use get_supported_version_path instead of get_supported_version when setting API path


gildub:
That's an easy one, well at least from my side :)
Please rebase so we can move forward.
Thanks again!

dhague:
Rebased and fixed hounds

gildub:
Thanks!

PR #98, opened 2016-05-06
Remove fog-xml dependency

voelzmo:
This seems to be unused and pulls in nokogiri. Unit tests run smoothly without fog-xml. Getting rid of nokogiri is always a good thing, so here we go.

Please double-check in case the units don't cover essential functionality.

cc @dhague, @Ladas, @seanhandley, @mdarby, @jjasghar


dhague:
Looks good to me, and I can't think of anywhere in OpenStack where XML is used.

PR #99, opened 2016-05-09
Identity class creation refactor

gildub:
This is step 1 towards identity v3 as the default version.
Effectively since Openstack Kilo, Keystone engine is by default version 3 while
still providing service to API v2.0.

This is one step closer to a configuration where the endpoints for v3 don't have
to be created along the default v2.0 which are not deprecated yet.

### DIFF
--- a/lib/fog/openstack/identity.rb
+++ b/lib/fog/openstack/identity.rb
@@ -14,16 +14,21 @@ module Fog
                  :current_user, :current_user_id, :current_tenant, :openstack_cache_ttl,
                  :provider, :openstack_identity_prefix, :openstack_endpoint_path_matches
 
-      # Fog::Identity::OpenStack.new() will return a Fog::Identity::OpenStack::V2 or a Fog::Identity::OpenStack::V3,
-      #  depending on whether the auth URL is for an OpenStack Identity V2 or V3 API endpoint
+      # Fog::Identity::OpenStack.new() will return a Fog::Identity::OpenStack::V3 by default
       def self.new(args = {})
-        if self.inspect == 'Fog::Identity::OpenStack'
-          identity = super
-          config = identity.config
-          service = identity.v3? ? Fog::Identity::OpenStack::V3.new(config) : Fog::Identity::OpenStack::V2.new(config)
-        else
-          service = Fog::Service.new(args)
+        version = '3'
+        url = Fog.credentials[:openstack_auth_url] || args[:openstack_auth_url]
+        if url
+          uri = URI(url)
+          version = '2.0' if uri.path =~ /v2\.0/
         end
+
+        service = case version
+                  when '2.0'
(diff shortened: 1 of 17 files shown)


lib/fog/openstack/identity.rb
```diff
@@ -35,14 +39,6 @@ def initialize(options = {})
           @openstack_identity_prefix = options[:openstack_identity_prefix]
```
dhague: With the v3? method removed, @openstack_identity_prefix is no longer used

gildub: Yeah, more likely, I haven't done a full clean-up yet. Good catch, I'll remove it. Thanks

gildub: That said @openstack_identity_prefix can still be used, although I don't have a use case handy.
Although it's not tested so that's bad!
To remove it would require a deprecation approach as it's used in many file

dhague: `options[:openstack_identity_prefix]` is used widely, but not `@openstack_identity_prefix` - that is only used in the Mock class in this file (and is there only for the v3? function) and is completely unused as a result of this commit.

tests/openstack/requests/image/image_tests.rb
```diff
@@ -1,5 +1,6 @@
 Shindo.tests('Fog::Image[:openstack] | image requests', ['openstack']) do
-  openstack = Fog::Identity[:openstack]
+  openstack = Fog::Identity::OpenStack::V2.new
```
Ladas: I think this will not work when you will run the tests against actutal env, with MOCK=false

Fog::Identity[:openstack] gets actualy a working identity, with credentials from ~/.fog

gildub:
Looks good. Just wondering about some examples/docs?

lib/fog/openstack/identity.rb
```diff
@@ -14,16 +14,21 @@ class OpenStack < Fog::Service
...
+      # Fog::Identity::OpenStack.new() will return a Fog::Identity::OpenStack::V3 by default
       def self.new(args = {})
-        if self.inspect == 'Fog::Identity::OpenStack'
-          identity = super
-          config = identity.config
-          service = identity.v3? ? Fog::Identity::OpenStack::V3.new(config) : Fog::Identity::OpenStack::V2.new(config)
-        else
-          service = Fog::Service.new(args)
+        version = '3'
+        url = Fog.credentials[:openstack_auth_url] || args[:openstack_auth_url]
```
Ladas: @gildub @dhague doh!, Fog.credentials[:openstack_auth_url]  should not be really here?

That means it always takes version of the API from my ~/.fog but then other params from the request.

Ladas: for the Fog.credentials, there should a separate path, where it is used for all params. But only if we don't provide it via args.

PR #100, opened 2016-05-09
Compute API: Volume attachments

dhague:


lib/fog/openstack/requests/compute/list_volume_attachments.rb
```diff
@@ -0,0 +1,31 @@
...
+            path:     format('servers/%s/os-volume_attachments', server_id)
+          )
+        end
+      end
+
+      class Mock
+        def list_volume_attachments(server_id)
+          Excon::Response.new(
+            body: {
+              volumeAttachments: [{
```
dhague: Given that we only support Ruby >= 1.9.x, I believe we should be using the new syntax in place of hash rockets

gildub: +1 
Really annoying to constantly going back and forth with the syntax.

lib/fog/openstack/requests/compute/list_volume_attachments.rb
```diff
@@ -0,0 +1,31 @@
...
+                device:     '/dev/vdd',
+                serverId:   server_id,
+                id:         '24011ca7-9937-41e4-b19b-141307d1b656',
```
gildub: +1
Ok, let's make a change to the Style/HashSyntax through another PR

PR #101, opened 2016-05-09
Compute API: Fix typo, minor bugfix, correct 'evacuate server' request body

dhague:


lib/fog/openstack/models/compute/server.rb
```diff
@@ -117,11 +117,19 @@ def reload
...
-          addresses.values.flatten.map{|x| x['addr']}
+          addresses ? addresses.values.flatten.collect { |x| x['addr'] } : []
         end
 
         def floating_ip_addresses
-          all_floating=addresses.values.flatten.select{ |data| data["OS-EXT-IPS:type"]=="floating" }.map{|addr| addr["addr"] }
+          all_floating = if addresses
+                           addresses.values.flatten.select { |data|
+                             data["OS-EXT-IPS:type"] == "floating"
+                           }.collect { |addr|
```
gildub: -1
Hound is right here, the block is hard to read, please use do...end over multiple lines.
Also you might have to use a proper if conditional block.

gildub:
Just a statement to layout differently, follow the hounds.

dhague:
Closed & reopened PR to trigger a rebuild

PR #102, opened 2016-05-09
Compute API: List availability zones

dhague:


gildub:
Easy and I'm not going to bother with the single quotes (we have so much legacy anyway).
Thanks

PR #103, opened 2016-05-09
Compute API: Update & delete flavor metadata

dhague:


lib/fog/openstack/requests/compute/delete_flavor_metadata.rb
```diff
@@ -0,0 +1,30 @@
...
+          response.headers = {
+            "X-Compute-Request-Id" => "req-fdc6f99e-55a2-4ab1-8904-0892753828cf",
+            "Content-Type"         => "application/json",
+            "Content-Length"       => "356",
+            "Date"                 => Date.new
+          }
+          response.body = nil
+          response
+        end
+      end # mock
```
gildub: I understand there are some legacy files using the `end # blah` but I believe this is not a good practice.
My stand is that any good IDE/text editor can collapse blocks easily if needed but more importantly this create a coupling between the comment and the code which in turn makes any changes difficult. Same for all three below.

lib/fog/openstack/requests/compute/update_flavor_metadata.rb
```diff
@@ -0,0 +1,33 @@
...
+            "Content-Length"       => "356",
+            "Date"                 => Date.new
+          }
+          response.body = {key => value}
+          response
+        end
+      end # mock
+    end # openstack
+  end # compute
+end # fog
```
gildub: Same as above,, no need for those end of block comments.

gildub:
Please follow the hounds in order to remove usage of `self.` when not needed.
Also I commented about the end of block `end # blah` usage, which I believe is wrong.
Thanks

lib/fog/openstack/compute.rb
```diff
@@ -227,73 +229,75 @@ class Mock
...
+                'injected_file_path_bytes'    => 256,
+                'injected_files'              => 5,
+                'metadata_items'              => 128,
+                'floating_ips'                => 10,
+                'instances'                   => 10,
+                'key_pairs'                   => 10,
+                'gigabytes'                   => 5000,
+                'volumes'                     => 10,
+                'cores'                       => 20,
+                'ram'                         => 51200
```
dhague: I do not plan to fix this one. #107 should see it go away anyway.

gildub:
Thanks!

PR #106, opened 2016-05-10
Replacing Shindo by minitest/spec - Step 1

gildub:
Per fog/fog#1266 and fog/fog#2630, Minitest is the new testing framework to be
used by fog projects.

This is the first step which:
- Adds /test
- Migrates /tests/openstack/*.rb from to /test/openstack/
- Adds minitest Rake rule
- Add minitest entry to Travis scripts

The following steps to:
- Revisit some of the describe/it texts
- Continue the migration of /tests
- Merge /test with /spec and Rake rules minitest with spec
- Remove /tests and Rake rule test


.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: ruby
 sudo: false
-script: bundle exec rake test
+script:
+  - bundle exec rake test
+  - bundle exec rake minitest
```
dhague: You could have both goals on one line, so:
script: bundle exec rake test minitest

gildub:
I think Travis needs a shell wrapper to be able to apply the command to various context (ruby versions)

Ladas:
looks good, ignoring hound fails which for the tests, would be actually uglier when fixed.

PR #107, opened 2016-05-10
Relax NumericLiterals rubocop rule to allow longer numbers 

dhague:


gildub:
Thanks, that's needed!!!

PR #108, opened 2016-05-11
implement network rbac policies

MaPoCaCe:
To access identity service for projects in network spec some adjustments to the shared_context was needed. And env `USE_VCR=true` now also works together with the `OS_` env present.

@dhague : FYI


dhague:
Fixed some more Hounds & rebased into a single commit

gildub:
Looks good. Just wondering about some examples/docs?

PR #109, opened 2016-05-11
Add NFV service

Ladas:
Add NFV service requests and models. Adding also request and
models tests and documentation.

### DIFF
--- a/lib/fog/openstack/models/collection.rb
+++ b/lib/fog/openstack/models/collection.rb
@@ -34,7 +34,10 @@ module Fog
       def get(uuid)
         raise Fog::OpenStack::Errors::InterfaceNotImplemented.new('Method :get is not implemented')
       end
-      alias_method :find_by_id, :get
+
+      def find_by_id(uuid)
+        get(uuid)
+      end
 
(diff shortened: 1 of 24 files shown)


Ladas:
damn, hound just got crazy and it stopped using our rubocop.yml

lib/fog/openstack/models/collection.rb
```diff
@@ -34,7 +34,10 @@ def summary(options = {})
       def get(uuid)
         raise Fog::OpenStack::Errors::InterfaceNotImplemented.new('Method :get is not implemented')
       end
-      alias_method :find_by_id, :get
+
+      def find_by_id(uuid)
+        get(uuid)
+      end
```
dhague: What does this achieve that alias_method didn't?

Ladas: gah, seems like this bug was here quite a while. Alias works weird with inheritance, so it always points to the above method, not the one implemented in the subclass. The things we discover with model tests. :-)

dhague: In which case: good catch!

PR #110, opened 2016-05-13
network doc and example

MaPoCaCe:
as suggested in https://github.com/fog/fog-openstack/pull/108

@dhague, @gildub


gildub:
That's really appreciated, thank you!

PR #116, opened 2016-05-17
Move docs/examples to root directory.

seanhandley:
Seems a better place, it's more visible from the repo root level on Github.


.gitignore
```diff
@@ -7,4 +7,5 @@
 /pkg/
 /spec/reports/
 /tmp/
-gemfiles/Gemfile-1.9.lock
\ No newline at end of file
+gemfiles/Gemfile-1.9.lock
+.DS_Store
```
gildub: Some clean up needed please.

gildub:
Yeah that's needed since the split from the core.
Just address the comment so we can merge.
Thanks

.gitignore
seanhandley: I added this because OSX Finder does this with annoying regularity.

I'll put it in a separate PR.

seanhandley:
Done. See #119

gildub:
Travis fails, one of the environment fails to install vcr 3.0.2, Rubygem version needs to be bumped up (>=2.5.2)

PR #118, opened 2016-05-19
Shindo to Minitest - Step 2

gildub:
Migrates baremetal, introspection

Also moves /test/openstack/* to /test/ 
Effectively /test/openstack seems to be coming from the *old day* of fog before the split ;)


gildub:
This one is replaced by PR #128 which is independent from the other "Shindo to Minitests ones".

PR #126, opened 2016-05-25
Shindo to Minitest - Common

gildub:
Moves test/openstack/*.rb to test/ and adjust Rakefile accordingly
Requires fog-core 1.40 (which adds match_schema assertion/expectation


gildub:
This is a common clean base for the incoming "Shindo to Minitest" sequel.
This touches only the new Minitest set (test/) and has no impact therefore I merge it so the following "Shindo to Minitests" Pull Requests can be submitted.

PR #127, opened 2016-05-25
Shindo to Minitest - Common part 2

gildub:
Adds `require` stanza in `test/test_helper.rb` to be able to use
match_schema assertion/expectation


gildub:
This is the second and last port for the common base needed for the incoming "Shindo to Minitest" sequel.
This impacts only the new Minitest set (test/). I merge directly so the following "Shindo to Minitests" Pull Requests can be submitted.

PR #128, opened 2016-05-25
Shindo to Minitest: baremetal, introspection

gildub:
Migrates baremetal, introspection from Shindo to Minitest
Note: We're not using subfolder 'openstack' folder anymore


test/requests/baremetal/chassis_tests.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    it "#list_chassis" do
+      @baremetal.list_chassis.body.must_match_schema('chassis' => [@chassis_format])
+    end
+
+    it "#list_chassis_detailed" do
+      @baremetal.list_chassis_detailed.body.must_match_schema('chassis' => [@detailed_chassis_format])
+    end
+
+    before do
```
petrblaho: Imho `before` block should be defined in the beginning of appropriate `describe` to be more readable.

test/requests/baremetal/chassis_tests.rb
```diff
@@ -0,0 +1,61 @@
...
+      'links'       => Array
+    }
+  end
+
+  describe "success" do
+    it "#list_chassis" do
+      @baremetal.list_chassis.body.must_match_schema('chassis' => [@chassis_format])
+    end
+
+    it "#list_chassis" do
```
petrblaho: Duplicated test as on line 25

test/requests/baremetal/node_tests.rb
```diff
@@ -0,0 +1,93 @@
...
+    it "#list_nodes" do
+      @baremetal.list_nodes.body.must_match_schema('nodes' => [@node_format])
+    end
+
+    it "#list_nodes_detailed" do
+      @baremetal.list_nodes_detailed.body.
+        must_match_schema('nodes' => [@detailed_node_format])
+    end
+
+    before do
```
petrblaho: Similar note as before - I would like to see `before` defined in the beginning of `describe` block.

test/requests/baremetal/port_tests.rb
```diff
@@ -0,0 +1,63 @@
...
+  describe "success" do
+    it "#list_ports" do
+      @baremetal.list_ports.body.must_match_schema('ports' => [@port_format])
+    end
+
+    it "#list_ports_detailed" do
+      @baremetal.list_ports_detailed.body.must_match_schema('ports' => [@detailed_port_format])
+    end
+
+    before do
```
petrblaho: The same as above.

petrblaho:
This maybe just my POV (and I am influenced by rspec heavily) but I do not like structure of specs like this:
```
describe "success" do
  it "#list_chassis" do
    @baremetal.list_chassis.body.must_match_schema('chassis' => [@chassis_format])
  end
end
```

I like this to be defined more like:
```
describe "#list_chassis" do
  describe "body" do
    it "matches schema" do
      @baremetal.list_chassis.body.must_match_schema('chassis' => [@chassis_format])
    end
  end
end
```
because this describes exactly what is tested (if `body` attribute of result of `#list_chassis` call `matches schema`) and using `--verbose` option you can read exactly from the output what failed.

But tests works and are readable so apart from notes with `before` block inline I am fine with that.

test/requests/baremetal/chassis_tests.rb
gildub: Fixed. Thanks!

test/requests/baremetal/chassis_tests.rb
gildub: Most of the tests in the requests section have a "success" and a "failure" context. This is why I've used the same approach and shared that under the top level "describe". Meanwhile this test doesn't have the "failure" context yet. That said I'm happy to regroup them for now.

test/requests/baremetal/node_tests.rb
gildub: This was a mistake, thanks!

gildub:
@petrblaho, I agree with you regarding the structure in order to make the test as clear as possible or at least more human friendly/readable. Meanwhile, please remember this task is to migrate everything from Shindo to Minitest which involves over 200 files total. So as you said the tests work that's the initial goal, now following commits to polish the describes.

PR #129, opened 2016-05-25
Shindo to Minitest: identity, image, planning

gildub:
Migrates identity, image and planning from Shindo to Minitest


gildub:
This one has been replaced with #131 and #132

PR #130, opened 2016-05-25
Shindo to Minitest - Common part 3

gildub:
Fix test identity leaks causing random errors
    
When Fog.credentials is altered by a test (or before) it must be restored at the end (or after) of the test.
    
This fixes authenticate_tests.rb and identity_version_tests.rb accordingly. It now avoids side effects cause some of the random test during testing.

### DIFF
--- a/test/authenticate_tests.rb
+++ b/test/authenticate_tests.rb
@@ -103,17 +103,21 @@ describe "OpenStack authentication" do
   end
 
   it "validates token" do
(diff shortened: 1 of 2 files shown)


gildub:
This is the third part for the common base needed for the incoming "Shindo to Minitest" sequel.
This impacts only the new Minitest set (test/) so I merge it myself directly so the following "Shindo to Minitests" Pull Requests can be submitted.

PR #131, opened 2016-05-25
Shindo to Minitest: identity

gildub:
Migrates identity from Shindo to Minitest


aufi:
Runned with ```bundle exec rake minitest```, works and looks good. Just minor note, I prefer using let(:something) instead of instance variables in tests personally, but if it is a convention here, no problem.

test/models/identity/roles_tests.rb
```diff
@@ -0,0 +1,28 @@
...
+  describe "success" do
+    it "#all" do
+      @roles.all.must_be_kind_of Fog::Identity::OpenStack::V2::Roles
+    end
+
+    it "#get" do
+      @roles.get(@roles.first.id).body.wont_be_nil
+    end
+
+    after do
```
petrblaho: I would move this `after` block in the beginning of appropriate `describe` block or maybe even one level upper here to the same level as `before` can be found.

petrblaho:
Apart from inline comment w/r/t `after` block LGTM even I prefer using `let` as @aufi mentioned.

test/models/identity/roles_tests.rb
gildub: I agree. Fixed it.

gildub:
@aufi, @petrblaho 
I agree with both of you regarding using #let instead of instance variables.
Meanwhle all those existing instance variables are coming from Shindo. Please understand the migration is an heavy task. So we initially want to pass all test and then having a polish phase for the cosmetic.

PR #132, opened 2016-05-25
Shindo to Minitest: image, planning

gildub:
Migrates image and planning from Shindo to Minitest


test/requests/image/image_tests.rb
```diff
@@ -0,0 +1,151 @@
...
+    it "#list_public_images" do
+      Fog::Image[:openstack].list_public_images.body.must_match_schema('images' => [@image_format])
+    end
+
+    it "#list_public_images_detailed" do
+      Fog::Image[:openstack].list_public_images_detailed.body.
+        must_match_schema('images' => [@detailed_image_format])
+    end
+
+    before do
```
petrblaho: I would move this in the beginning of `describe` block.

test/requests/image/image_tests.rb
```diff
@@ -0,0 +1,151 @@
...
+      end
+
+      @instance = Fog::Image[:openstack].create_image(image_attributes).body
+    end
+
+    it "#create_image" do
+      @instance.must_match_schema('image' => @detailed_image_format)
+    end
+
+    after do
```
petrblaho: The same as with `before` block above.

petrblaho:
Apart from two `before`/`after` blocks location (see inline notes) looks good to me.

test/requests/image/image_tests.rb
gildub: Done. Thanks

test/requests/image/image_tests.rb
gildub: Done. Thanks

PR #133, opened 2016-05-25
Shindo to Minitest: nfv

gildub:
Migrates nfv from Shindo to Minitest

Adds test/helpers/nfv_helper.rb to share common nvf data


test/helpers/nfv_helper.rb
```diff
@@ -0,0 +1,27 @@
+def set_nfv_data
```
petrblaho: I do not like this method - what it does is that it creates a few instance variables, which is clear side effect and it is hidden from the user of this method.

I suppose that splitting this method into 3 functions with no creation of instance variables but just returning proper values would be more clean solution. 

Or just return it from this function so it can be used as `@nfv, @vnf_data, @auth = set_nfv_data` which is only a little longer call and gives user of this function control over where are data stored without unseen side effects.

test/models/nfv/vnf_tests.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    it "#destroy" do
+      sleep(10) unless Fog.mocking?
+
+      @vnfs.destroy
+      @vnfd.destroy.must_equal true
+    end
+
+    after do
```
petrblaho: I would move `after` block just after `before` block so we can see what is happening before _and_ after each test without need to read through all tests.

petrblaho:
Other than inline comments LTGM.

test/helpers/nfv_helper.rb
gildub: I agree. Done

test/models/nfv/vnf_tests.rb
gildub: Done. Thanks

PR #134, opened 2016-05-25
Shindo to Minitest: orchestration, metering, volume

gildub:
Migrates metering, orchestration and volume


test/requests/volume/quota_tests.rb
```diff
@@ -0,0 +1,40 @@
...
+    it "#get_quota_defaults" do
+      @volume.get_quota_defaults(@tenant_id).body.
+        must_match_schema('quota_set' => @quota_set_format)
+    end
+
+    it "#get_quota" do
+      @quota.must_match_schema(@quota_set_format)
+    end
+
+    it "updates quota" do
```
petrblaho: This test has two asserts - line 33 and line 37. When first one fails the second one will never be run (and it can fail too). This introduces delay in developers workflow - solve the first one failing just to discover the second one failing too.

Another issue with more than one assert per `it` is connected to what I wrote at https://github.com/fog/fog-openstack/pull/128#issuecomment-222156071 - it is the structure of `describe`/`it` and naming of them. Should one of these asserts fail I will see
`Fog::Volume[:openstack] | quota requests::success#test_0003_updates quota [/home/pblaho/workspace/fog/fog-openstack/test/requests/volume/quota_tests.rb:37]:` and with information about expected and actual result. That line suggests where the failing assert is located (line 37 here) but it tells me that it is `Fog::Volume[:openstack] | quota requests::success#test_0003_updates quota` test failing.

With proper naming and structure of `describe`/`it` blocks and only one assert per test you can immediately see all failed tests and exactly what test failed and developer who knows the codebase can just know what the issue could be.

If there is a need to test two subsequent quota updates I would just create two `it` blocks, make them share the same `describe` and `before` block to setup tests, and the first would just call `#update_quota` once and assert result and the second would call it twice and assert result.

petrblaho:
If we can solve two asserts in one `it` I am happy to say that LGTM.

test/requests/volume/quota_tests.rb
gildub: Thanks for the detailed explanation. I agree the break down of describe/it can/must be improved for better readability. Again this is legacy code and the initial target is to port it to Minitest.
Regarding the specific "updates quota" I effectively don't see the point of asserting #update_quota twice. Therefore I remove the second one.

PR #135, opened 2016-05-25
Shindo to Minitest: common4

gildub:
Adds a migrated version of collection_helper.rb & model_helper.rb
helpers to Minitest in /test/helpers

Also, adds a UNIT_TESTS_CLEAN flag to skip broken test


gildub:
This is the fourth part for the common base needed for the incoming "Shindo to Minitest" sequel.
This impacts only the new Minitest set (test/) so I merged it directly for the following "Shindo to Minitests" Pull Requests to be able to be submitted.

PR #139, opened 2016-05-26
Compute create_server issue with volume

gildub:
Found a bug in create_server which is not detected by Shindo
because of helper method overlap


dhague:
+1 LGTM

seanhandley:
Good catch 👍

PR #144, opened 2016-06-07
various fixes for host aggregate calls

sseago:
This fixes a few problems with host aggregates:
1) add_host call was succeeding but throwing an exception
2) remove_host call was succeeding but throwing an exception
3) create was not setting the availability_zone attribute
4) update was not working properly


seanhandley:
👍

seanhandley:
Thanks @sseago

PR #145, opened 2016-06-07
Mock volume's status when attach and detach perform

KevinLoiseau:
Hi guys !
Just a little PR for updating volume's status when that was attached or detached.
Thanks


dhague:
LGTM - sorry it took so long, I've been away from Fog for a little while

PR #146, opened 2016-06-22
networking token renewal is not possible on token auth

Carthaca:
When I authenticate with auth-token, obviously I cannot get a new token on expiration.
It was nevertheless tried and resulted in the confusing error with status 400: `Expecting to find id or name in user - the server could not comply with the request since it is either malformed or otherwise incorrect. The client is assumed to be in error.` instead of correctly passing though the 401: `Authentication required` indicating the consumer of the library that a new token needs to be provided.

I will tackle the other services, too.


Ladas:
Makes sense to me. Thank you for the contribution. :+1:

PR #147, opened 2016-06-24
Token renewal fix

Carthaca:
If the initial authentication happened with a token, there is no chance of renewing the token - so we should properly fail.

Following https://github.com/fog/fog-openstack/pull/146, where the fix is exercised on one service I take the chance to refactor the `request` method into the `core` module.

Please note the following details:
- module `common` was the role-model
- I moved the path changes out into the `set_api_path` (newly introduced for metering service)
- and the NotFound error class into `self.not_found_class` (fallback to compute. baremetal, orchestration and planning should introduce their own in the future I propose)
- storage service had the additional request option `parse_json`, that needs to be removed later, in favor of the request param `:raw_body` I propose


dhague:
LGTM - nice work.

PR #149, opened 2016-06-28
Add workflow service

rwsu:
Adds support for Mistral.


seanhandley:
This is great work @rwsu - thanks a lot ❤️

PR #152, opened 2016-07-01
Update Glance examples

jockey10:
Add example for downloading image from Glance. Additionally, combine the
image creation from URL and download examples, as both need to use the
upload_data call.


Ladas:
@jockey10 thank you

PR #153, opened 2016-07-04
maintain @path in local variable _path

devtdeng:
@path instance variable will be overwritten in case authenticate is invoked, keep the value in to local variable _path so that it won't lose original value.

### DIFF
--- a/lib/fog/openstack/core.rb
+++ b/lib/fog/openstack/core.rb
@@ -78,13 +78,14 @@ module Fog
       def request(params, parse_json = true)
         retried = false
         begin
+          _path=@path
           response = @connection.request(params.merge(
                                            :headers => {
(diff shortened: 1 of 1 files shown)


lib/fog/openstack/core.rb
```diff
@@ -78,13 +78,14 @@ def reload
       def request(params, parse_json = true)
         retried = false
         begin
+          _path=@path
```
Ladas: ruby doesn't usually use the python _ and __ prefixes. Just put there 'path'. _ in ruby marks it's not used, e.g. unused param.

Might be that the error is actually in the authenticate? What is being placed in the path, when authenticate is called? It should be always set to the right value.

seanhandley:

PR #160, opened 2016-07-15
Change token based v3 authentication

dhague:
First try to validate given token. If not successful, use old re-authenticate logic for compatibility reasons.
The validation should keep the numbers of issued tokens in keystone much smaller, because a token will be reused instead of a new one is created.

Re-record VCR tests accordingly, and also allow recordings can go to different folders


Ladas:
@dhague do you think you could separate the style changes and the actual change? I got lost, it's too many changes. :-)

gildub:
@dhague,  are you still working on this?

dhague:
@gildub I was waiting for someone to approve the PR - I don't think there's anything more to do, although it looks like I should rebase and resubmit, which I will do shortly.

dhague:
@gildub Looks good for merging now - would you do the honours?

lib/fog/openstack.rb
```diff
@@ -129,7 +129,7 @@ def self.authenticate(options, connection_options = {})
       when /v2(\.\d+)?/
         authenticate_v2(options, connection_options)
       when /v3(\.\d+)?/
-        authenticate_v3(options, connection_options)
+        validate_or_authenticate_v3(options, connection_options)
```
gildub: I believe there should be 2 different methods: #validate and #authenticate_v3.
And add any logic at this level if needed.

lib/fog/openstack.rb
```diff
@@ -227,30 +226,30 @@ def self.authenticate_v2(options, connection_options = {})
...
-        :server_management_url    => management_url,
-        :token                    => body['access']['token']['id'],
-        :expires                  => body['access']['token']['expires'],
-        :current_user_id          => body['access']['user']['id'],
-        :unscoped_token           => options[:unscoped_token]
+        :server_management_url => management_url,
+        :token => body['access']['token']['id'],
+        :expires => body['access']['token']['expires'],
+        :current_user_id => body['access']['user']['id'],
+        :unscoped_token => options[:unscoped_token]
```
gildub: This was actually right hash style indentation, at least according to rubocop.

lib/fog/openstack.rb
```diff
@@ -331,17 +330,96 @@ def self.authenticate_v3(options, connection_options = {})
...
-          :current_user_id          => body['token']['user']['id'],
-          :unscoped_token           => options[:unscoped_token]
+        :user => body['token']['user']['name'],
+        :tenant => tenant,
+        :identity_public_endpoint => identity_url,
+        :server_management_url => management_url,
+        :token => token,
+        :expires => body['token']['expires_at'],
+        :current_user_id => body['token']['user']['id'],
+        :unscoped_token => options[:unscoped_token]
```
gildub: Same here, no need to change well aligned hash.

lib/fog/openstack.rb
```diff
@@ -331,17 +330,96 @@ def self.authenticate_v3(options, connection_options = {})
...
+
+      identity_service_type = options[:openstack_identity_service_type]
+      openstack_region = options[:openstack_region]
+
+      service = get_service_v3(body, service_type, service_name, openstack_region, options)
+      identity_service = get_service_v3(body, identity_service_type, nil, nil, :openstack_endpoint_path_matches => %r{/v3}) if identity_service_type
+
+      # Services not found  -> unscoped token
+      # Fallback to standard authenticate to be compatible
+      if service
```
gildub: Could the service validation be in its own method to separate it from the token validation itself?
Although, when OpenStack returns a token there are some services attach to it through the catalog. Normally that's what should be used for service validation, now if that doesn't fit fog-openstack it's a different story.

seanhandley:
The complexity of this is mind boggling 😢

seanhandley:
Are you still pursuing this @dhague or shall I close?

dhague:
@seanhandley I've not looked at this in a very long time, and things seem to be ticking along nicely without it...

I would close it.

Issue #49, opened 2016-03-10
Create/list/delete server groups and assign VMs into them


plribeiro3000:
Hello,

Does somebody know whether OpenStack provider implments creation/deletion/listing of server groups as well as asignment of VMs into them.
Basically I am interested in creation of virtual machine in a certain server group:
```
nova boot  --image ubuntu --flavor m1.small --hint group=aa8a69f4-2567-47c6-a28c-2934a8c7959c vm1
```

Thanks

Original issue opened by @pdilung at fog/fog#3656

danleyden:
setting the server group when creating a VM works (when calling `servers.create` pass in `os_scheduler_hints: { group: server_group_id }`)

creating, listing and deleting of server groups is not yet supported, but I have a patch that adds such (no tests or examples yet - see 7acbef0286c386493ad6db566a5ccca82745a259 - but proven out locally)

plribeiro3000:
@danleyden Would you open a PR with your changes?

danleyden:
PR was opened 2 weeks ago, no comments thus far

plribeiro3000:
@danleyden could you link the PR here please?
Thanks

danleyden:
https://github.com/fog/fog-openstack/pull/161

PR #161, opened 2016-07-15
supporting CRUD for server groups issue #49

danleyden:


lib/fog/openstack/models/compute/server_group.rb
```diff
@@ -0,0 +1,22 @@
...
+        attribute :name
+        attribute :policies, :type => :array
+        attribute :members
+
+        VALID_SERVER_GROUP_POLICIES = ['affinity', 'anti-affinity']
+
+        def self.validate_server_group_policy(policy)
+          fail ArgumentError, "#{policy} is an invalid policy... must use one of #{VALID_SERVER_GROUP_POLICIES.join(', ')}" \
+            unless VALID_SERVER_GROUP_POLICIES.include? policy
+          true
```
plribeiro3000: This is confusing imo. You return true if it is valid but raise an exception if its wrong. What do you think of returning true or false only?

lib/fog/openstack/requests/compute/create_server.rb
```diff
@@ -169,6 +169,12 @@ def create_server(name, image_ref, flavor_ref, options = {})
 
           self.data[:last_modified][:servers][server_id] = Time.now
           self.data[:servers][server_id] = mock_data
+
+          if options['os:scheduler_hints'] && options['os:scheduler_hints']['group']
+            grp = self.data[:server_groups][options['os:scheduler_hints']['group']]
```
plribeiro3000: `grp` stands for group right?
Can you be more explicitamente and use `group` instead?

lib/fog/openstack/requests/compute/delete_server.rb
```diff
@@ -22,6 +22,8 @@ def delete_server(server_id)
               self.data[:last_modified][:servers].delete(server_id)
               self.data[:servers].delete(server_id)
               response.status = 204
+              group_id, _ = data[:server_groups].find { |_id, grp| grp[:members].include?(server_id) }
```
plribeiro3000: Could you please take care of this @houndci-bot comment?

lib/fog/openstack/models/compute/server_group.rb
danleyden: That could return true / false, but then the places where that is called should raise the ArgumentError unless it was true (less efficient due to having two checks inline and also duplication of that check / raise)

lib/fog/openstack/requests/compute/list_server_groups.rb
```diff
@@ -0,0 +1,28 @@
...
+            :path    => 'os-server-groups',
+            :query   => options
+          )
+        end
+      end
+
+      class Mock
+        def list_server_groups(_options = {})
+          groups = data[:server_groups].map do |id, grp|
+            grp.merge('id' => id, 'project_id' => 'test-project', 'user_id' => 'test-user')
```
plribeiro3000: Use `group` here as well please.

plribeiro3000:
@danleyden looking good besides some small comments.
Thank you for opening this PR!

lib/fog/openstack/models/compute/server_group.rb
```diff
@@ -0,0 +1,22 @@
...
+module Fog
+  module Compute
+    class OpenStack
+      class ServerGroup < Fog::OpenStack::Model
+        identity  :id
+        attribute :name
+        attribute :policies, :type => :array
+        attribute :members
+
+        VALID_SERVER_GROUP_POLICIES = ['affinity', 'anti-affinity']
```
plribeiro3000: Could you fix this as well please?

lib/fog/openstack/models/compute/server_group.rb
plribeiro3000: Well. fair enough.

plribeiro3000:
@danleyden Could you take a look at the broken build as well please?

plribeiro3000:
@danleyden I can see there is more instances named `grp` but that should not be a blocker. Does this code currently work? Have you tried it?

danleyden:
I have successfully run this code against a deployment of IceHouse. I don't have access to a deployment of anything more recent.

However, from looking at the API documentation, it *should* work with newer versions.

plribeiro3000:
👍

plribeiro3000:
Sadly, i don't have permissions to merge.
Let me try to get it.

plribeiro3000:
Done. Thanks for your contribution @danleyden !

ShimiTaNaka:
Any idea when it will be out?

plribeiro3000:
@ShimiTaNaka I don't have permissions for `fog-openstack` on rubygems.

cc/ @geemus Can you cut a new release or give me access to do so?

geemus:
released 0.1.10

PR #162, opened 2016-07-19
Fix stack create action

Ladas:
Fix stck create action, getting not found on template, when
stack has not been created yet. Brought by:
https://github.com/fog/fog-openstack/pull/147


gildub:
Thanks

PR #164, opened 2016-07-20
remove default_options in #save

foreverqihe:
when using save to create stacks with "template_url", it causes error because default_opotions have a template: null. Heat API takes template over template_url even if it is null. This default options is used for those deprecated methods such as #template=() etc. But for #save(), we do not need the default_options because it prevents us from setting our now option values. @mdarby @geemus @tokengeek @icco @dhague, @Ladas, @seanhandley and @jjasghar Do you guys have time to approve this pull request?

### DIFF
--- a/lib/fog/openstack/models/orchestration/stack.rb
+++ b/lib/fog/openstack/models/orchestration/stack.rb
@@ -15,9 +15,9 @@ module Fog
 
         def save(options={})
           if persisted?
-            service.update_stack(self, default_options.merge(options)).body['stack']
+            service.update_stack(self, options).body['stack']
           else
-            service.stacks.create(default_options.merge(options))
+            service.stacks.create(options)
           end
         end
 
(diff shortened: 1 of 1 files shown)


mdarby:
LGTM

Ladas:
that was a quick merge, I believe this is not right, you are breaking backwards compatibility by not sending the default params

I did this fix, which is just fixing the error we brought https://github.com/fog/fog-openstack/pull/162

foreverqihe:
@Ladas, the issue is not "not_found_class". The issue is when you create stack using "template_url", the "template" parameter in default_options are shadowing the template_url, but template is nil. Heat API complains no template specified.

Ladas:
@foreverqihe right, so probably it's better to just handle that case. You patch also fixed the not_found_error, cause by getting the default template, the whole stack create failed.

Could you just handle the template url and template case? E.g. by checking which one of those is not nil and setting only that. That way we should not break a backwards compatibility.

foreverqihe:
@Ladas yes, sure. Could you merge #162 first? Because default_options throw error without it.

Ladas:
@foreverqihe #162 was merged

foreverqihe:
@Ladas yes, thanks. And I created #165, could you approve it?

geemus:
Thanks!

PR #166, opened 2016-07-25
add routes option to router update to add extra routes

Carthaca:
and align mock


Carthaca:
resolved conflicts in https://github.com/fog/fog-openstack/pull/180

PR #167, opened 2016-07-25
fallback for empty addresses

Carthaca:
introducing a guard


Carthaca:
will do a new PR with a cherrypick

Carthaca:
see https://github.com/fog/fog-openstack/pull/179

PR #168, opened 2016-08-02
Initial DNS Designate implementation

Carthaca:
This implements basic v1 + v2 list domains/zones and quota get + update

There will come more :)


lib/fog/openstack/requests/dns_v1/get_quota.rb
```diff
@@ -0,0 +1,26 @@
...
+            response = Excon::Response.new
+            response.status = 200
+            response.body = data[:quota_updated] || data[:quota]
+            response
+          end
+        end
+      end
+    end
+  end
+end
```
plribeiro3000: The path of this file should be `lib/fog/dns/openstack/v1/requests/get_quota.rb`.

This follows the pattern we have defined right now in other providers. Could you move this file and other ones that do not follow this pattern please?

plribeiro3000:
@Carthaca Great Work here. Just a small comment and i believe we are good.
Also, why did you implement `v1` and `v2` and no just `v2` altogether?

lib/fog/openstack/requests/dns_v1/get_quota.rb
Carthaca: Thanks for having a look - I will move the files.

Carthaca:
I started implementing with designate stable/mitaka, where we don't have v2/quotas.

Afterwards we decided to go to the designate master branch to get those features - but I feel keeping v1 is worth for all that don't run bleeding edge designate :)

lib/fog/openstack/requests/dns_v1/get_quota.rb
plribeiro3000: 👍

plribeiro3000:
@Carthaca Sure. Sounds good. Thanks for sharing this.

Carthaca:
With all that changes already, I don't feel like correcting all the hound complaints, too.

Carthaca:
🐕  😃

Issue #78, opened 2016-04-11
Can not reload glance image model


kei-yamazaki:
When create glance image model via glance v2 api, following error occurs.

```
E, [2016-04-11T07:51:38.359236 #5466] ERROR -- : identity is required for this operation (ArgumentError)
```

I think this issue related to [this code](https://github.com/fog/fog-openstack/blob/118a99282e1225db8735f5547280788ede97a09e/lib/fog/openstack/models/image_v2/image.rb#L51-L51).

gildub:
I've not been to the bottom of it yet, but I don't think reload should be used.
Also a file path must be passed to upload_data, and affiliated requests, not the data itself:
```ruby
aki.upload_data "#{extract_path}/cirros-0.3.0-x86_64-vmlinuz"
```
And the image is effectively loaded and active:

```ruby
p image.get_image_by_id(aki.id).body["status"]
"active"
```

kei-yamazaki:
@gildub, You are right, but the script that I have posted is just test.
If I just get the state of glance image, I think that using the image.get_image_by_id is better.

The script can upload glance image via glance v2 api, but can not reload itself.

kei-yamazaki:
I use `reload` in common method that checks resources, such as vm and volume.

gildub:
@kei-yamazaki, sure but reload is not implemented for Image class.
That said the error provided is not really helpful.

Ladas:
I noticed this while ago. I use reload e.g. for checking state of uploaded image, so it's nice to have, but there is a workaround. I might fix this, somewhere in the future, unless somebody else will step in. :-)

kei-yamazaki:
@Ladas, I hope to fix this problem.

Ladas:
@kei-yamazaki cool :-)

kei-yamazaki:
@Ladas, Oh, sorry my English is incorrect.
I hope that you fix this problem!

Ladas:
@kei-yamazaki ah, ok :-) though not sure, when I will have time to do this :-)

PR #170, opened 2016-08-03
Make image_v2 models more consistent with Fog API

ljfranklin:
We would like our project to support both v1 and v2 Image APIs as some OpenStack deployments might only support one or the other. However, this gem makes a couple breaking changes in the v2 Image implementation:
- Images cannot be reloaded
- `image.save` always calls create even if the image already exists
- `image.update` must be passed a hash, which breaks the pattern of setting instance variables then calling `image.save`

We also made a couple cosmetic changes, such as allowing `identity_v3_spec.rb` to use the VCR variables when `use_recorded` is true.

After this patch is merged, https://github.com/fog/fog-openstack/issues/78 can be closed.

Let us know if something looks amiss and we'll take another crack at it. Thanks!


ljfranklin:
We'll add a commit to make Hound Bot happy in the next couple hours.

ljfranklin:
Looks like the Travis failures were unrelated to any changes we made, it was unable to install the gem `tins` on ruby 1.9.3. Might think about dropping tests for 1.9.3 as it has end End-Of-Life since February 23, 2015 :)

lib/fog/openstack/models/image_v2/images.rb
```diff
@@ -18,9 +18,9 @@ def summary(options = {})
           end
 
           def find_by_id(id)
-            image_hash = service.get_image_by_id(id).body
-            Fog::Image::OpenStack::V2::Image.new(
-                image_hash.merge(:service => service))
+            result = all.find { |image| image.id == id }
```
voelzmo: Why did you switch here from getting an individual image to listing all images? The API should return something similar to your error message (`itemNotFound (404)`) in case the image with that id doesn't exist.

lib/fog/openstack/models/image_v2/image.rb
```diff
@@ -66,27 +71,44 @@ def create
...
               service.update_image(id, json_patch).body)
             self
           end
 
+          # This overrides the behaviour of Fog::OpenStack::Model::save() which tries to be clever and
+          #  assumes save=update if an ID is present - but Image V2 allows ID to specified on creation
+          def save
+            create
+          rescue StandardError => e
+            raise e unless e.response && e.response.body && e.response.body.match(/Attribute .* is read-only./)
```
voelzmo: Relying on messages in the response body is very brittle, but I assume you already found that you cannot rely on return codes here to figure out what's going on at the server?

lib/fog/openstack/models/image_v2/images.rb
ljfranklin: Fog core's `reload` function strangely requires a collection to be set on the model. Creating a model with `Image.new` would throw "undefined method 'get' on NilClass" when reload was called as [`collection` was nil]((https://github.com/fog/fog-core/blob/f6ef6c3d9458bc235be57100acdced264e85e52a/lib/fog/core/model.rb#L43). I'm not sure whether there's a better why to avoid the nil pointer exception short of using `all.find`.

lib/fog/openstack/models/image_v2/image.rb
ljfranklin: I would have expected the API to return a 409 Conflict if the resource already exists, in fact it will under some circumstances. But in our test cases for Image V2, calling `create` if the ID already exists instead returned 403 "Attribute 'self' is read-only". We'll open an issue on the OpenStack API to look into this further.

ljfranklin: Submitted a bug report to Glance here: https://bugs.launchpad.net/glance/+bug/1611154

ljfranklin: Bug report aside, it's probably better anyway to check for the presence of an attribute like `self` to know whether to call `create` or `update`. Less brittle and potentially avoids an extra HTTP call. We'll fix that tomorrow: https://www.pivotaltracker.com/story/show/127953859.

ljfranklin: This is fixed in https://github.com/fog/fog-openstack/pull/170/commits/49de1abcbb033b7973ccf4a6eec5f2d9df50a8ee

lib/fog/openstack/models/image_v2/images.rb
dhague: Why not use the pattern that is common throughout fog-openstack? Something like this (for the Identity V2 tenant object, fog-openstack/lib/fog/identity/openstack/v2/models/tenants.rb):

```
          def find_by_id(id)
            cached_tenant = find { |tenant| tenant.id == id }
            return cached_tenant if cached_tenant
            tenant_hash = service.get_tenant(id).body['tenant']
            Fog::Identity::OpenStack::V2::Tenant.new(
              tenant_hash.merge(:service => service)
            )
          end
```

Carthaca: This would not do an actual reload (new query), but take the cached value.

I suggest the following:

```ruby
def find_by_id(id)
  new(service.get_image_by_id(id).body)
end
```

Carthaca:
The tins with ruby 1.9.3 issue should be solved with https://github.com/fog/fog-openstack/commit/4ef26991793df1d37b43a15b7613e2f6ae41e990 - so a rebase should do the trick ;)

lib/fog/openstack/models/image_v2/images.rb
ljfranklin: @Carthaca @dhague If we go back to the original implementation (`service.get_image_by_id(id).body`) instead of `all.find`, we get the following error when calling `reload`:

```
Fog::Image::OpenStack#test_0004_Reloads image state from the server:
NoMethodError: undefined method `get' for nil:NilClass
    /Library/Ruby/Gems/2.0.0/gems/fog-core-1.42.0/lib/fog/core/model.rb:43:in `reload'
    /Users/pivotal/workspace/fog-openstack/spec/image_v2_spec.rb:150:in `block (3 levels) in <top (required)>'
    /Library/Ruby/Gems/2.0.0/gems/vcr-3.0.3/lib/vcr/util/variable_args_block_caller.rb:9:in `call'
    /Library/Ruby/Gems/2.0.0/gems/vcr-3.0.3/lib/vcr/util/variable_args_block_caller.rb:9:in `call_block'
    /Library/Ruby/Gems/2.0.0/gems/vcr-3.0.3/lib/vcr.rb:189:in `use_cassette'
    /Users/pivotal/workspace/fog-openstack/spec/image_v2_spec.rb:125:in `block (2 levels) in <top (required)>'
```

This nil pointer is due to the base model having a nil `collection`: https://github.com/fog/fog-core/blob/f6ef6c3d9458bc235be57100acdced264e85e52a/lib/fog/core/model.rb#L43.

Any suggestions for an implementation that only fetches a single object from the server, but avoids this nil pointer exception?

ljfranklin:
@Carthaca just pushed the rebase from master. See our comment above about fixing the nil pointer exception on reload.

lib/fog/image/openstack/v2/models/images.rb
```diff
@@ -17,10 +17,9 @@ def summary(options = {})
           end
 
           def find_by_id(id)
-            image_hash = service.get_image_by_id(id).body
-            Fog::Image::OpenStack::V2::Image.new(
-              image_hash.merge(:service => service)
-            )
+            result = all.find { |image| image.id == id }
```
Carthaca: I can't find the comment on the outdated code anymore, so here it goes again: note the tiny difference that I'm not suggesting to initialize a new Image, but a collection:

```ruby
def find_by_id(id)
  new(service.get_image_by_id(id).body)
end
```

ljfranklin: @Carthaca whoops, misunderstood your original comment. Your fix worked perfectly, here's the [commit](https://github.com/fog/fog-openstack/pull/170/commits/bc4c198a4fdbd5180d03dde5bf286afea218556d)!

voelzmo:
@Carthaca, @dhague  Can we get this change merged or are there any additional issues you're seeing? A release with this change would remove a pretty ugly monkeypatch we currently do...
Thanks!

lib/fog/image/openstack/v2/models/image.rb
```diff
@@ -66,28 +71,45 @@ def create
...
+            end
+
             merge_attributes(
               service.update_image(id, json_patch).body
             )
             self
           end
 
+          # This overrides the behaviour of Fog::OpenStack::Model::save() which tries to be clever and
+          #  assumes save=update if an ID is present - but Image V2 allows ID to specified on creation
```
gildub: Did you meant "...allows ID to be specified on creation" ?

cunnie: Typo fixed in a82e2bb

cunnie:
Thanks @gildub,

- we fixed the typo (thanks!)
- squashed our commits into one self-contained commit
- rebased against master, resolving conflicts in `spec/shared_context.rb`

— @ljfranklin and @cunnie

ljfranklin:
I think we fixed all the Hound issues except for "Align the elements of a hash literal if they span more than one line." as I don't know what it means. We thought it wanted us to line up the `=>` symbols, but it still didn't like it.

Some gentle feedback about the Hound integration, it's crazy noisy. My inbox is showing 96 emails from Hound on this PR and counting. I'm guessing every other person that is subscribed to this PR got the same amount of emails and immediately muted the thread. I also have no way of knowing whether I fixed the style error until I push. We tried running `rake rubocop:auto_correct`, but this changed ~70 files, and we didn't feel comfortable pushing that many changes (it seemed that rubocop with autocorrect hadn't been run on the repo in a long time, if ever).

IMO it would be much easier / less noisy if you ran a Rubocop rake task in Travis which performed the same style validations. This would keep the noise in the thread to a minimum by giving the PR author a single notification that their PR failed the Travis build due to style shortcomings.

Carthaca:
Hi, it's noisy yes, but we have to take the pain right now 😄
To the alignments: you should go to the longest key, set exactly one space to the rocket `=>` and align the other rockets to that. Hth

lib/fog/image/openstack/v2/models/image.rb
```diff
@@ -162,6 +171,42 @@ def remove_tag(tag)
             requires :id
             service.remove_tag_from_image(id, tag)
           end
+
+          private
+
+          def build_update_json_patch(client_attributes, server_attributes)
```
Ladas: I am fine with this now, but d oyou think you could extract all below methods to some helper for PATCH, these methods will be reusable elsewhere.

ljfranklin: @Ladas as this PR almost 2 months old at this point, I really don't want to touch it again :) While removing duplicate code is nice, I don't think this blocks the merging of this PR.

voelzmo:
@Ladas @Carthaca @dhague @gildub Is there something additional necessary to get this merged? Seems to me that there weren't any roadblocks left. Wdyt?

Ladas:
:+1:

PR #175, opened 2016-08-11
DNS V2 crud zone + recordsets

Carthaca:
as promised some more dns v2 requests


lib/fog/dns/openstack/v2/requests/update_recordset.rb
```diff
@@ -0,0 +1,37 @@
+module Fog
+  module DNS
+    class OpenStack
+      class V2
+        class Real
+          def update_recordset(zone_id, id, options = {})
```
dhague: What happens with the options here? It looks like they are just thrown away...

dhague:
LGTM, although some VCR tests would also be nice

PR #178, opened 2016-08-15
[wip] Add Glance v2 image download Content-Range support

mansam:
This implements Content-range support for the Glance v2 image download API in the most basic way, maintaining the placeholder function signature that was already there.

@Ladas  This should probably be more user friendly, maybe with `offset` and `chunk_size` parameters instead of just taking the raw header value, but I wanted to get some opinions before doing anything fancy. Also, what is this project's preferred way of testing a change like this?

### DIFF
--- a/lib/fog/image/openstack/v2/requests/download_image.rb
+++ b/lib/fog/image/openstack/v2/requests/download_image.rb
@@ -3,12 +3,16 @@ module Fog
     class OpenStack
       class V2
         class Real
-          def download_image(image_id, _content_range = nil, params) # TODO: implement content range handling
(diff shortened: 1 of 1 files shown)


mansam:
A fix for the Glance bug which prevents this API from working correctly appears to be incoming. https://bugs.launchpad.net/glance/+bug/1618928

With that in mind, would it be possible to discuss the requirements to this or something like it merged? This simplistic implementation would do the job but I'd like to get feedback on what kind of API would be preferred.

mansam:
@ladas Any thoughts on using the "Connection: close" header in this API call to sidestep the Glance bug?

seanhandley:

PR #180, opened 2016-08-17
add routes option to router update to add extra routes

Carthaca:
and align mock

### DIFF
--- a/lib/fog/network/openstack/requests/update_router.rb
+++ b/lib/fog/network/openstack/requests/update_router.rb
@@ -11,7 +11,7 @@ module Fog
         def update_router(router_id, options = {})
           data = {'router' => {}}
 
-          [:name, :admin_state_up].each do |key|
+          [:name, :admin_state_up, :routes].each do |key|
             data['router'][key] = options[key] if options[key]
           end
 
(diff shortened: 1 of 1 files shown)


lib/fog/network/openstack/requests/update_router.rb
```diff
@@ -53,9 +53,11 @@ def update_router(router_id, options = {})
             if egi.kind_of?(Fog::Network::OpenStack::Network)
               Fog::Logger.deprecation "Passing a model objects into options[:external_gateway_info] is deprecated. \
               Please pass  external external gateway as follows options[:external_gateway_info] = { :network_id => NETWORK_ID }]"
-              router[:external_gateway_info] = {:network_id => egi.id}
-            else egi.kind_of?(Hash) && egi[:network_id]
-                 router[:external_gateway_info] = egi
+              router[:external_gateway_info] = { :network_id => egi.id }
+            elsif egi.is_a?(Hash)
+              router[:external_gateway_info] = egi
```
dhague: does it not matter if egi has a :network_id?

Carthaca: Have a look at the `Real` class, please - I made sure that the mock is handling it the same way.

Contentwise: that was removed to allow clearing the gateway by setting the network_id to `nil` in https://github.com/fog/fog-openstack/commit/6427cad70deeab9004276ac30c867d2088440c1e

dhague:
LGTM

PR #184, opened 2016-09-03
Fix introspection service lookup

rwsu:
Upstream, the service is registered as "baremetal-introspection"
as the service type name. Setting the service type name manually
produced resulted in an error in set_api_path because it could not
find a supported version.

This patch changes the type from "introspection" to
"baremetal-introspection". It also explicitly sets the api path to
v1 because the version reported by the introspection api is 1.6
which leads Fog::OpenStack.get_supported_version to throw an error.

### DIFF
--- a/lib/fog/introspection/openstack.rb
+++ b/lib/fog/introspection/openstack.rb
@@ -80,7 +80,7 @@ module Fog
         def initialize(options = {})
           initialize_identity options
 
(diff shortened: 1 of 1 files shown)


tzumainn:
:+1: 

@Ladas, think this is mergeable/releaseable?

petrblaho:
:+1:

Ladas:
@rwsu hopefully this will not break the backwards compatibility? I do not have a means to test, so merging now

PR #186, opened 2016-09-03
fix example of connection options v2

demetrodon:


Ladas:
@demetrodon thank you :+1:

PR #187, opened 2016-09-05
[DNS] add models for zone + recordset incl spec

Carthaca:
add header options for admin to query for all_projects or a specific other project

in https://github.com/sapcc/fog-openstack/blob/53b66cb021a6254ea9a69c13dd348d2b6ab0b8b0/lib/fog/dns/openstack/v2/requests/list_recordsets.rb#L7-L17 I would prefer to do the breaking change instead of that deprecation workaround - what do you think? we only offer this DNS API for 4 weeks - most likely it is not consumed, yet or the adaption still fresh to adjust for the change 😄

### DIFF
--- a/lib/fog/dns/openstack/v2/models/zone.rb
+++ b/lib/fog/dns/openstack/v2/models/zone.rb
@@ -1,0 +1,50 @@
+require 'fog/openstack/models/model'
+
+module Fog
+  module DNS
+    class OpenStack
+      class V2
+        class Zone < Fog::OpenStack::Model
+          identity :id
+
+          attribute :name
+          attribute :email
+          attribute :pool_id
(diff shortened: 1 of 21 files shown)


lib/fog/dns/openstack/v2/models/zone.rb
```diff
@@ -0,0 +1,50 @@
...
+          attribute :project_id
+          attribute :serial
+          attribute :status
+          attribute :action
+          attribute :masters
+          attribute :version
+          attribute :links
+          attribute :created_at
+          attribute :transfered_at
+
```
dhague: Maybe add a comment to say that the following attributes are optional

dhague:
LGTM

PR #188, opened 2016-09-07
Add a request for "Cancel stack update"

foreverqihe:
Hello ,

  I created a cancel_update request for orchestration since it is missing and I need to use it. I tested it with Heat and it works:  i.e. the stack is rolled back during the update. 

  Could you merge it? Thank you in advance. @Ladas @mdarby 

Regards,
HeQi


Ladas:
@foreverqihe could you add just a simple spec to https://github.com/foreverqihe/fog-openstack/blob/7f8ad0462ebf35a1ffbc960df9e8a92b8a901756/test/requests/orchestration/stack_tests.rb ? So it at least goes through that code. And maybe a small doc here https://github.com/foreverqihe/fog-openstack/blob/7f8ad0462ebf35a1ffbc960df9e8a92b8a901756/docs/orchestration.md

thank you

foreverqihe:
@Ladas Done, could you have a look?

Ladas:
@foreverqihe awesome, thank you :-)

foreverqihe:
@Ladas Thanks! When can we make a release( so I don't need to use github as gem source)?

Issue #183, opened 2016-08-31
Unecessary requires


plribeiro3000:
There are some requires at `fog/openstack` that are not required because the constants are already being loaded by `autoload`.
This also breaks the purpose of using autoload to decrease the loading time when requiring the main `fog` gem.

Please remove those requires [here](https://github.com/fog/fog-openstack/blob/master/lib/fog/openstack.rb#L5-L20)

PR #191, opened 2016-09-08
Fix workflow service lookup

rwsu:
Adds missing require. In some usages, this error occurs:
uninitialized constant Fog::Workflow::OpenStack::V2

### DIFF
--- a/lib/fog/openstack.rb
+++ b/lib/fog/openstack.rb
@@ -18,6 +18,7 @@ require 'fog/planning/openstack'
 require 'fog/storage/openstack'
 require 'fog/volume/openstack/v1'
 require 'fog/volume/openstack/v2'
+require 'fog/workflow/openstack/v2'
 
 module Fog
   module Compute


plribeiro3000:
@rwsu I've filled another issue regarding those requires in #183 
Could you please change this so it will use autoload only please?

Thanks!

rwsu:
@plribeiro3000 autoload doesn't appear to working in my case. Any ideas why? I'm using the fog-openstack gem from a web app. https://gist.github.com/rwsu/b53b2ab3cc7d7c7cbfc35cf53a22a88f

plribeiro3000:
@rwsu How did you define your autoload?

rwsu:
@plribeiro3000 I am not sure how autoload is supposed to be used. Can you provide an example?
Without this patch I would have to add the following to a class in rails for things to work correctly.

require 'fog/openstack'
require 'fog/workflow/openstack/v2'

plribeiro3000:
@rwsu You have to change [lines 68-70](https://github.com/rwsu/fog-openstack/blob/6d6459ad46189538ff54a970d0fdb0d94701c22d/lib/fog/openstack.rb#L68-L70) and include the new class as below:

```ruby
module Workflow
  module OpenStack
    autoload :V1, File.expand_path('../workflow/openstack/v1', __FILE__)
    autoload :V2, File.expand_path('../workflow/openstack/v2', __FILE__)
  end
end
```

Your current setup does not work because `V2` is defined inside `V1`. This will have to be changed as well.

gildub:
@plribeiro3000,

I agree with the fact that autoload should be used upfront and issue #183 needs to be addressed meanwhile I don't think that's the point of this PR.
I'm tempted to merge that one and have a separate patch for fixing the missing autoload.

PR #196, opened 2016-09-22
Fix MIME issues

charlesdunbar:
This adds the fixes found in
https://github.com/fog/fog-aws/commit/13bda13933711c7d3efa00bcf1e913e053aa34e6
that come up due to changing some gem dependencies around.

See https://github.com/fog/fog-core/issues/171 and
https://github.com/fog/fog-core/pull/170

Without this, trying to use "put_object" for storage results in:

```
NameError: uninitialized constant Fog::Storage::MIME
```

### DIFF
--- a/lib/fog/storage/openstack.rb
+++ b/lib/fog/storage/openstack.rb
@@ -43,7 +43,22 @@ module Fog
       request :post_set_meta_temp_url_key
       request :public_url
 
+      module Utils
+        def require_mime_types
+          # Use mime/types/columnar if available, for reduced memory usage
+          require 'mime/types/columnar'
+          rescue LoadError
+            begin
+              require 'mime/types'
+            rescue LoadError
+              Fog::Logger.warning("'mime-types' missing, please install and try again.")
+              exit(1)
+            end
+          end
+      end
+
       class Mock
+        include Utils
         def self.data
           @data ||= Hash.new do |hash, key|
             hash[key] = {}
@@ -55,6 +70,7 @@ module Fog
         end
 
         def initialize(options = {})
+          require_mime_types
           @openstack_api_key = options[:openstack_api_key]
           @openstack_username = options[:openstack_username]
           @path = '/v1/AUTH_1234'
@@ -80,6 +96,7 @@ module Fog
       end
 
       class Real
+        include Utils
         include Fog::OpenStack::Core
 
         def self.not_found_class
@@ -87,6 +104,7 @@ module Fog
         end
 
         def initialize(options = {})
+          require_mime_types
           initialize_identity options
 
           @openstack_service_type           = options[:openstack_service_type] || ['object-store']


gildub:
Please fix the hound. The first level of rescue doesn't need `begin/end` since it's wrapped by `def/end`.
As mentioned fog-core/issue/171, the lazy require makes sense meanwhile I don't understand the loading failure sequence:
If `mime/types/columnar` loading fails then it's probably because `mime-types` is missing anyway.
Am I missing something?

charlesdunbar:
Hound fixed.

https://github.com/mime-types/ruby-mime-types/#columnar-store mentions that columnar wasn't introduced until mime-types 2.6, so I assume the loading sequence is to allow for older versions of mime-types to work.

autoload is deprecated - https://bugs.ruby-lang.org/issues/5653

lib/fog/storage/openstack.rb
```diff
@@ -43,7 +43,22 @@ class OpenStack < Fog::Service
       request :post_set_meta_temp_url_key
       request :public_url
 
+      module Utils
+        def require_mime_types
+          # Use mime/types/columnar if available, for reduced memory usage
+          require 'mime/types/columnar'
+          rescue LoadError
```
Ladas: could you fix the alignment?

charlesdunbar: Whoops - missed this one, should be fixed now.

lib/fog/storage/openstack.rb
```diff
@@ -43,7 +43,22 @@ class OpenStack < Fog::Service
...
+            begin
+              require 'mime/types'
+            rescue LoadError
+              Fog::Logger.warning("'mime-types' missing, please install and try again.")
+              exit(1)
+            end
+          end
```
charlesdunbar: I can't satisfy every piece of the hound.

module
  def
    rescue
      rescue
      end
    end
end

With 4 openings, and the implicit "begin" in def, I only need 3 endings.  I can't align them to all of them.  I can align to the def or the first rescue, what's your preference?

seanhandley:

Kiemes:
@seanhandley @charlesdunbar @Ladas: Could we try to get this PR merged? This definitely still is an issue caused by fog-core which delegates the loading of `mime/types` to its users.
Currently also all users of fog-openstack need to require `mime/types` themselves as we do not require it here.

As far as I can see, the only objection for merging was the formatting of the code because the hound complained.

Could the repo owners just merge the commit and fix the alignment to their liking?

gildub:
@Kiemes 

Could you please details the error you're getting. 

I might be completely missing the issue, but since mime/types are not used at all with fog-openstack and fog-core has the require optional I don't understand while we should be including it at all.

Also the gem is only declared for test/dev.

Kiemes:
@gildub sure.

In https://github.com/fog/fog-openstack/blob/master/lib/fog/storage/openstack/requests/put_object.rb#L23 `Fog::Storage.parse_data` is called which is provided by fog-core.
This one does `get_content_type` which does `MIME::Types.of` https://github.com/fog/fog-core/blob/master/lib/fog/storage.rb#L69

So when we are using fog-openstack and saving a file in our blobstore, we get an error saying `NameError: uninitialized constant Fog::Storage::MIME`. That's how we identified this issue/PR.

gildub:
@Kiemes 

Thank you for providing those crystal clear details 👍 

I agree with you about fog-core to require the gem since it's overwriting it.

Let's poke the fog-core folks again...

lib/fog/storage/openstack.rb
gildub: I believe the issue is that when using implicit `begin` then `rescue` should be aligned with `def`. For instance:
```ruby
module M
  def foo
    ...
  rescue
    begin
    rescue
    end
  end
end
```

gildub:
We effectively need this in the meantime support for 1.9.3 is dropped.

gildub:
@charlesdunbar, 

Could you please check my comment, that could also make the hound happy.

Also bringing back the `autoload` approach on the front burner: I believe [1] is outdated, autoload is still in use in Ruby 2.3.1 and is not going to disappear, it seems the initial issues and fears which existed at the time of autoload not being safe have been addressed. So I would re-consider using it. What do you think?

[1] https://bugs.ruby-lang.org/issues/5653

YaEvan:
When I use the fog-openstack in rails5.1.4 and ruby2.3.4, I follow the storage manual run:
file = directory.files.create :key => 'space.jpg', :body => File.open("/home/evan/space.jpg")

The exception is 
NameError: uninitialized constant Fog::Storage::MIME
from /home/evan/.rvm/gems/ruby-2.3.4@fog-openstack/gems/fog-core-1.45.0/lib/fog/storage.rb:69:in `get_content_type'

I fix this problem by adding `gem 'mime-types', require: 'mime/types/full'` in gemfile.
I wonder if there is any problem with this usage?

voelzmo:
*bump* @gildub @charlesdunbar: in the meanwhile, ruby 1.9 support has been dropped, so can we re-consider this?

gildub:
@voelzmo, fog-core 1.45 has been released so this effectively seems to be obsolete.

Kiemes:
@gildub if you want to make sure that people do not run in this issue, you might consider bumping fog-core in the gemspec (https://github.com/fog/fog-openstack/blob/9eff6ae4316edddf21c69cad538127a0a96002a8/fog-openstack.gemspec#L24)

gildub:
@kiemes, thanks, just waiting for broken tests to be fixed: https://github.com/fog/fog-openstack/commit/2b32ce8b3d000b20d196601ddbda048d9a463554

voelzmo:
cool, thanks!

gildub:
There is no reason to hold this one any longer because of aforementioned issue with local test which not directly related, as long as CI works! We should have done that earlier, anyway, merging now. Thanks

PR #198, opened 2016-09-23
Replaces requires with autoload

gildub:
Fix https://github.com/fog/fog-openstack/issues/183

Reordered the autoloaded by alphabetical order.

Note: Left planning to be done later
because its using a non consistent path and name.


Ladas:
@gildub great :+1:

Issue #199, opened 2016-09-28
Neutron: Need to list extensions


gildub:
Please see http://developer.openstack.org/api-ref/networking/v2/index.html for details.

PR #201, opened 2016-09-29
Enhancement: adds neutron extensions

gildub:
Includes
Request: list_extensions, get_extension
Models: extension
Collection: extensions
Unit tests for above

Reference: https://github.com/fog/fog-openstack/issues/199


Ladas:
@gildub awesome, thank you

PR #205, opened 2016-09-30
Manila Shares

Carthaca:
Initial manila implementation.

I plan to add snapshots and quota soon.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -74,6 +74,7 @@ These services can be explored in further depth in the following documents:
 * [NFV](nfv.md)
 * [Orchestration](orchestration.md)
 * [Planning](planning.md)
+* [Shared File System](shared_file_system.md)
 * [Storage (Swift)](storage.md)
 * [Workflow](workflow.md)
 
--- a/lib/fog/openstack.rb
+++ b/lib/fog/openstack.rb
@@ -67,6 +67,10 @@ module Fog
     autoload :OpenStack, File.expand_path('../orchestration/openstack', __FILE__)
   end
 
+  module SharedFileSystem
+    autoload :OpenStack, File.expand_path('../shared_file_system/openstack', __FILE__)
+  end
+
   module Storage
     autoload :OpenStack, File.expand_path('../storage/openstack', __FILE__)
   end
@@ -91,21 +95,22 @@ module Fog
   module OpenStack
     extend Fog::Provider
 
-    service(:compute,       'Compute')
-    service(:image,         'Image')
-    service(:identity,      'Identity')
-    service(:network,       'Network')
-    service(:storage,       'Storage')
-    service(:volume,        'Volume')
-    service(:metering,      'Metering')
-    service(:orchestration, 'Orchestration')
-    service(:nfv,           'NFV')
-    service(:baremetal,     'Baremetal')
-    service(:planning,      'Planning')
-    service(:introspection, 'Introspection')
-    service(:monitoring,    'Monitoring')
-    service(:workflow,      'Workflow')
-    service(:dns,           'DNS')
+    service(:compute,            'Compute')
+    service(:image,              'Image')
+    service(:identity,           'Identity')
+    service(:network,            'Network')
+    service(:storage,            'Storage')
+    service(:volume,             'Volume')
+    service(:metering,           'Metering')
+    service(:orchestration,      'Orchestration')
+    service(:nfv,                'NFV')
+    service(:baremetal,          'Baremetal')
+    service(:planning,           'Planning')
+    service(:introspection,      'Introspection')
+    service(:monitoring,         'Monitoring')
+    service(:workflow,           'Workflow')
+    service(:dns,                'DNS')
+    service(:shared_file_system, 'SharedFileSystem')
 
     @token_cache = {}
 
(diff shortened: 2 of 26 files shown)


Carthaca:
I disabled the `FrozenStringLiteralComment` cop - I don't think it is planned to go for ruby 3.0 in the near future

gildub:
@Carthaca, LGTM.
We can merge once you have added the doco and examples.
Thanks

Carthaca:
added docu + example and squashed the commits :)

gildub:
The intent to add the resource for the documention is noble but I don't think that should belong to this patch. Could you please keep here only the documentation related to the Manila shares?

Carthaca:
done - I will open a new PR with the documentation consolidation

docs/getting_started.md
```diff
@@ -68,14 +68,15 @@ Fog::OpenStack.services
 These services can be explored in further depth in the following documents:
 
 * [Compute](compute.md)
+* [Introspection](introspection.md)
+* [Metering](metering.md)
+* [Network](network.md)
+* [NFV](nfv.md)
+* [Orchestration](orchestration.md)
+* [Planning](planning.md)
+* [Share (Manila)](share.md)
```
gildub: Documentation is effectively behind, thank you for doing this.
Meanwhile could you please make those 'common' changes into a separate Pull Request?
So we can keep track of changes.

docs/getting_started.md
```diff
@@ -68,14 +68,15 @@ Fog::OpenStack.services
...
 ## Additional Resources
-* [fog.io](http://fog.io)
-* [Fog rdoc](http://rubydoc.info/gems/fog)
-* [Fog Github repo](https://github.com/fog/fog)
-* [Release Notes](https://github.com/fog/fog/blob/master/changelog.txt)
-
-
-## Support and Feedback
-Your feedback is appreciated! If you have specific issues with the **fog** SDK, you should file an [issue via Github](https://github.com/fog/fog/issues).
+[resources and feedback](shared/resources.md)
```
gildub: Same as above, please submit a separate PR for those changes not related to your initial Manila commit.
Also, I think, `shared` is confusing I would opt for `common` but it's only my view of it. :)

docs/share.md
```diff
@@ -0,0 +1,82 @@
+# Share (Manila)
```
gildub: I believe the generic component name is "Shared Filed System" not just "Share". 
Please change the name accordingly everywhere it needs to in order to reflect the full name, including in the file: `shared_file_system.md`. Thanks

docs/shared/connection_params.md
```diff
@@ -0,0 +1,45 @@
+### Optional Connection Parameters
```
gildub: As mentioned earlier, this one to go within a separate PR.
Also not keen on `shared` name.

examples/share/basics.rb
```diff
@@ -0,0 +1,59 @@
+require 'fog/openstack'
```
gildub: As above, rename examples/shared_file_system/basics.rb

lib/fog/network/openstack/requests/create_network.rb
```diff
@@ -69,13 +69,13 @@ def create_network(options = {})
...
-            'shared'                => options[:shared],
+            'shared'                => options[:shared] || false,
             'subnets'               => [],
             'status'                => 'ACTIVE',
-            'admin_state_up'        => options[:admin_state_up],
+            'admin_state_up'        => options[:admin_state_up] || false,
             'tenant_id'             => options[:tenant_id],
             'qos_policy_id'         => options[:qos_policy_id],
-            'port_security_enabled' => options[:port_security_enabled]
+            'port_security_enabled' => options[:port_security_enabled] || false
```
gildub: Same as above. That's better but not in this PR.

lib/fog/openstack.rb
```diff
@@ -106,6 +110,7 @@ module OpenStack
     service(:monitoring,    'Monitoring')
     service(:workflow,      'Workflow')
     service(:dns,           'DNS')
+    service(:share,         'Share')
```
gildub: I actually initially missed that, I think the name could be `:shared_file_system`.

lib/fog/share/openstack.rb
```diff
@@ -0,0 +1,239 @@
+module Fog
+  module Share
```
gildub: To match a full generic name I think the module name `Share` would be something like SharedFileSystem. What do you think?

lib/fog/share/openstack/models/network.rb
```diff
@@ -0,0 +1,42 @@
+require 'fog/openstack/models/model'
+
+module Fog
+  module Share
```
gildub: Same as above, ShareFileSystem or so

Carthaca:
Thanks for having a look.

To the network changes: I simply needed them to get the tests green :) You are right, that this should be kept separate, same with the common documentation changes. I will do that.

To the name: I used the service catalog as orientation, the service type of manila is 'share'. I like it better, because it is shorter than 'shared_filed_system'.

Carthaca:
I am done with moving out the not manila related stuff.

Do we need a third opinion for the name or can we agree on 'share'?

Carthaca:
Okay, it's now or never ^^

"Shared File System" is more specific.

gildub:
Thank you for this great work!

PR #209, opened 2016-10-06
Allows security_groups to update

gildub:
Adds update call to request and corresponding model

### DIFF
--- a/lib/fog/network/openstack/models/security_group.rb
+++ b/lib/fog/network/openstack/models/security_group.rb
@@ -22,9 +22,14 @@ module Fog
         end
 
         def save
-          raise Fog::Errors::Error, 'Resaving an existing object may create a duplicate' if persisted?
           merge_attributes(service.create_security_group(attributes).body['security_group'])
-          true
+          self
+        end
+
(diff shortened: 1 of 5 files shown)


lib/fog/network/openstack/models/security_group.rb
```diff
@@ -22,9 +22,14 @@ def security_group_rules
         end
 
         def save
-          raise Fog::Errors::Error, 'Resaving an existing object may create a duplicate' if persisted?
           merge_attributes(service.create_security_group(attributes).body['security_group'])
-          true
+          self
+        end
+
+        def update
```
Ladas: maybe we could refactor this later to provider the standard interface of the base class create/update and save methods

Ladas:
the, the hound is not deleting the non valid comments

Ladas:
@gildub thanks :-)

lib/fog/network/openstack/models/security_group.rb
gildub: :+1:  Yes, I thought about that too. In my todo list...

PR #215, opened 2016-10-12
disable cop intended for ruby 3.0 migration

Carthaca:
I don't think it is planned to go for ruby 3.0 in the near future

moved out of #205

@gildub FYI


gildub:
Yes apparently available only since Ruby 2.3 it won't be default until Ruby 3+. No need to get the hounds hunting use until then! Thanks

PR #216, opened 2016-10-12
Stack update correction

dpansare:
Hi,

The stack update feature retrieves the current stack template and then merges user's template/ template url to it. However if the user is trying to update with a template url, it gets ignored since the template(self) is always present. 
This change is to consider user's template/template url if it's provided instead of the current stack's template. 
Could you please merge it? Thank you in advance. @Ladas @mdarby

Thanks
Deepali

### DIFF
--- a/lib/fog/orchestration/openstack/models/stack.rb
+++ b/lib/fog/orchestration/openstack/models/stack.rb
@@ -14,7 +14,11 @@ module Fog
 
         def save(options = {})
           if persisted?
-            service.update_stack(self, default_options.merge(options)).body['stack']
+            stack_default_options = default_options
+            if (options.key?(:template_url))
+              stack_default_options.delete(:template)
+            end
+            service.update_stack(self, stack_default_options.merge(options)).body['stack']
           else
             service.stacks.create(default_options.merge(options))
           end


dpansare:
@dhague, @Ladas, @seanhandley, @mdarby and @jjasghar Could I please request a review and merge? 

Thanks
Deepali

dhague:
@dpansare The change looks good, but could you rebase & squash so that it doesn't take 7 commits to change 5 lines of code?

lib/fog/orchestration/openstack/models/stack.rb
```diff
@@ -14,7 +14,11 @@ class Stack < Fog::OpenStack::Model
 
         def save(options = {})
           if persisted?
-            service.update_stack(self, default_options.merge(options)).body['stack']
+            stack_default_options = default_options
+            if [:template,:template_url].any? { |e| options.key?(e) }
```
dhague: I think you only need to check for `:template_url` here. If `:template` is in the options then it will overwrite `:template` in `default_options` when the `merge(options)` is done in line 21, so there is no need to specifically remove it from `default_options`.

dpansare:
@dhague Did rebase and squash and changed code to check only template_url. Could you please merge if found ok? 

Thanks

PR #220, opened 2016-10-27
Remove redundant object_id attribute

gildub:
This ain't necessary and generates a pesky warning:
"lib/fog/core/attributes/default.rb:53: \
warning: redefining `object_id' may cause serious problems"


gildub:
@ladas, you're going to love it and not believe it at same time ;)

Ladas:
@gildub oh yeah, I've already hit similar. It's the problem with exposing all attributes as getters, sometime we hit internal names :-(

Ladas:
@gildub if we need that attribute somewhere, we can still get it with .attributes['object_id']

PR #221, opened 2016-10-31
Fix glance image upload example

jockey10:
This commit fixes incorrect logic for the glance image upload. The
"upload_data" method was incorrectly accessed from the glance object,
and not the handle to the created image.


gildub:
Good find. Thanks

Issue #222, opened 2016-11-01
Mock classes belong to test


gildub:
The various Service's Mock classes belong to the tests section, namely `test/`.

For more details, please see conversation in fog-core [1]. Although the work happening in fog-core should be completely transparent for fog-openstack, we might follow same approach (or not) of using an external gem (fog-mock-openstack) for keeping all Mock code. 

This could happen in stages, initially separating the Mock classes and related code from the source code.

[1] https://github.com/fog/fog-core/issues/198

PR #223, opened 2016-11-01
Moving Mock to test: part 1

gildub:
Just moves some Mock blocks into separate files under `test/lib`

https://github.com/fog/fog-openstack/issues/222


test/lib/fog/identity/openstack/v3.rb
```diff
@@ -0,0 +1,32 @@
...
+  module Identity
+    class OpenStack
+      class V3 < Fog::Service
+        class Mock
+          include Fog::OpenStack::Core
+          def initialize(options = {})
+          end
+        end
+
+        def self.get_api_version(uri, connection_options = {})
```
Carthaca: I believe that method is not needed here

gildub: You're right, thanks

gildub:
@Carthaca, @ladas, @aufi, would you mind finishing review. I really would like to move on with getting Mock out of the way, where it belongs in tests, so we can finish consolidate tests and that would help to make the Real class evolve.

plribeiro3000:
I'm a little late for this but wont this change break the feature of mocking?
Wont it be available only from within the tests of `fog-openstack` and not the tests of the system using `fog-openstack`?

gildub:
@plribeiro3000, 

Yes the validation of the feature along the feature would be moved to tests only. Because although Mock/Real classes pattern is interesting, and more likely made lot of sense in fog's early days, it's not a good practice and can easily be eplaced with Webmock, VCR and integrations tests and other equivalent.

As expressed earlier in the https://github.com/fog/fog-core/issues/198 discussion, the production code mixed up with test code doesn't guarantee a better SDK where the equivalent feature can be obtained and offer more flexibility and modularity.

plribeiro3000:
Hmmmm. I believe we have a problem here.
The extraction of the providers to its own repos and gems were meant to help the end user. So it does not have to load all proiders when it is going to use just one.
But in the end we still want to make a library there is as close as possible of an abstraction that works for many providers with little changes.
Mock is a feature that is all over the providers and we do have several users that use it. Removing it shouldn't be a matter that simple.
It affects all providers so we need to be on the same spot before we start dropping features, otherwise we will end up will several repos completely different from one another.

I truly believe this move was wrong in the sense that we did not had conclude the discussion neither you guys have communicated with the rest of the maintainers of the other repos about such change.

Can we hold this until we get to a conclusion please?

PS: I need your emails so i can register you on our slack to wrap this up.

plribeiro3000:
cc/ @geemus

geemus:
This is a tricky one. The hope historically had been that by providing mocks, we could simplify individual users usage because they wouldn't all individually have to maintain webmocks/vcr/etc. ie it was intended as an end-user feature, more than a internal testing facilitation tool. I certainly agree that the line has gotten blurry there and it gets used very heavily within our own tests also though.

@gildub If I understand correctly here, it sounds as though you are suggesting that we now distribute that responsibility to the user base? Perhaps we should tease this apart. I can certainly imagine moving to VCR/webmock for our own testing purposes, but the mock feature doesn't necessarily have to be dropped at the same time. Does that make sense? What do you think?

PR #224, opened 2016-11-01
Add Gnocchi metric service

aufi:
This PR adds a basic subset of features of Gnocchi OpenStack service http://gnocchi.xyz/

The Gnocchi service extracted part of Ceilometer functionality related to Metrics. It is a new service in OpenStack catalog.

### DIFF
--- a/examples/metric/basics.rb
+++ b/examples/metric/basics.rb
@@ -1,0 +1,42 @@
+require 'fog/openstack'
+require 'time'
+
+auth_url = "http://10.0.0.13:5000/v3/auth/tokens"
+username = 'admin'
+password = 'njXDF8bKr68RQsfbANvURzkmT'
+project  = 'admin'
+
+@connection_params = {
+    :openstack_auth_url     => auth_url,
+    :openstack_username     => username,
+    :openstack_api_key      => password,
+    :openstack_project_name => project,
+    :openstack_domain_id    => "default"
+}
+
+puts "### SERVICE CONNECTION ###"
+
+metric = Fog::Metric::OpenStack.new(@connection_params)
+
+p metric
+
+puts "### RESOURCES ###"
+
+p metric.list_resources
+
+p metric.resources.all(details: true)
+
+p metric.resources.find_by_id("3c6c53c9-25c1-4aca-984d-a20c1926b499")
+
+p metric.get_resource_metric_measures("d1f84147-d4ef-465e-a679-265df36918ed", "disk.ephemeral.size", start: 0, stop: Time.now.iso8601, granularity: 300).body
+
+
+puts "### METRICS ###"
+
+p metric.metrics.all
+
+p metric.metrics.find_by_id("7feff2ca-2edd-4ea5-96d7-2cc5262bb504")
+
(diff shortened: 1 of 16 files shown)


seanhandley:
Awesome work here 👍 

Curious why it's under the `Metric` namespace. My understanding was that Gnocchi can be used for metering in place of Ceilometer (which is due to be retired). See the Etherpad from the Barcelona Summit: https://etherpad.openstack.org/p/ocata-summit-telemetry-deprecating-ceilometer-api

aufi:
@seanhandley Thanks for comment! The main reason for a new namespace was that Gnocchi is a new service in Openstack Catalog (other Service Type than Metering). It can be moved under Metering, but I am not sure about its internal structure then. Would you have a recommendation about that?

seanhandley:
No @aufi - I'm not very familiar with Gnocchi but I know Ceilometer's agents can send data into Gnocchi's database.

Ladas:
@aufi @seanhandley Metric namespace sounds good to me, so far we've been consistent with naming the namespace according to keystone service name. So we should keep doing that. After some time we should delete the whole Metering namespace(1 or 2 releases?) , since it will not be supported anymore.

aufi:
@seanhandley @Ladas I am not sure why travis is failing. Is there a known issue with ```public_suffix gem``` on old ruby or jruby?

seanhandley:
Why is there a requirement for the `public_suffix` gem @aufi ? And can that functionality be implemented internally?

The Fog project likes to keep its dependency footprint as small as possible.

aufi:
@seanhandley public_suffix gem comes from webmock -> addressable -> public_suffix and it was updated so it requires ruby 2.0 or higher (after running bundle update). I tried lock its version to older one to resolve travis fail.

It would be nice remove it, but currently I dont have other idea how to resolve travis fail (or if it is not merge-blocker - no problem, can remove it).

seanhandley:
Ahh I see - thanks for clarifying @aufi. And webmock doesn't pin the version of public_suffix so it's free to be upgraded to a Ruby 2 only version...

Hmmm. This feels like it could be affecting the broader `fog` project so I'll need to dig around.

There's always this approach in the meantime:

```ruby
# fog-openstack.gemspec
if RUBY_VERSION < '2'
  spec.add_development_dependency 'public_suffix', '~> 1.5.3' 
end
```

seanhandley:
Given the age of Ruby 1.9, I think the project is keen to drop support fairly soon.

@Ladas we should also plan when to drop 1.9 support

aufi:
@seanhandley Makes sense, thanks! PR was updated.

Ladas:
@seanhandley I would say we could drop 1.9 soon, it's not being supported for a year now. Should we do a major release of fog-openstack for that?

aufi:
@Ladas Agree on dropping 1.9 in some time..

I added few tests, would it be merge-able now?

Ladas:
@aufi awesome, thank you :+1:

seanhandley:
@Ladas - probably worth a minor release to `0.2`. It's not a major rewrite from the code's perspective so I don't think it's worth bumping to `1.0`.

Might be nice to do a final `0.1.x` release with a deprecation warning, though?

Ladas:
@seanhandley yeah, deprecation warning would be nice

seanhandley:
See #229

PR #225, opened 2016-11-03
Keystone v3 examples

jockey10:
Some examples of using Keystone v3, based on the previous v2 examples. I've tested these against an OpenStack deployment with Keystone v3 configured, and a single 'Default' domain.


Ladas:
@jockey10 thank you

PR #238, opened 2016-11-25
Improve create server

scambra:
When I set port_id in nics, it fails to create, if I remove uuid from request, it works.

Also, I need some more parameters in block_device_mappings_v2: 
device_type, disk_bus, guest_format, virtual_name

### DIFF
--- a/lib/fog/compute/openstack/requests/create_server.rb
+++ b/lib/fog/compute/openstack/requests/create_server.rb
@@ -61,7 +61,7 @@ module Fog
             data['server']['block_device_mapping_v2'] = [block_device_mapping].flatten.collect do |mapping|
               entered_block_device_mapping = {}
               [:boot_index, :delete_on_termination, :destination_type, :device_name, :device_type, :disk_bus,
-               :guest_format, :source_type, :uuid, :virtual_name, :volume_size].each do |index|
+               :guest_format, :source_type, :uuid, :volume_size].each do |index|
                 entered_block_device_mapping[index.to_s] = mapping[index] if mapping.key?(index)
               end
               entered_block_device_mapping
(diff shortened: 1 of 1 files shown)


seanhandley:
Thanks @scambra 

Can you tell us some details about the OpenStack platform you're running this against?

i.e. Release level, Nova version etc

This code's very widely used in production so we need to understand why it works for some users and not others.

scambra:
# nova --version
2.17.0
# dpkg -l nova-common|grep nova
ii  nova-common                          1:2014.1.5-0ubuntu1.4

seanhandley:
I'll revisit this soon @scambra - December/January have been hard months with regard to spare time!

guigui89:
Hello, is this pull request going to be approved and merged? I would like to use this feature for my project.

Carthaca:
Hi, the first part lgtm.

But I don't see the device_type, disk_bus & virtual_name in the [api-ref](https://developer.openstack.org/api-ref/compute/?expanded=create-server-detail#create-server), can you point me to where these are listed as valid options, please?

guigui89:
Hello, I would only need the first part of this pull request (commit ref: db6d424dcbbb245f7b74bb7f9159ca8a21e804d0)

scambra:
I'm not using guest_format and virtual_name anymore, I don't remember what I was using for, but I'm still use device_type and disk_bus and they work. I found doc here for guest_format, device_type and disk_bus:

https://docs.openstack.org/developer/nova/block_device_mapping.html

If you agree, I can remove virtual_name from pull request

scambra:
Here is doc about virtual_name, maybe it's old parameter which has been removed

Carthaca:
`virtual_name` is legacy, it was changed 4 years ago: https://github.com/openstack/nova/commit/d581ed04a584fdd83df78e43366a3617b154a626

Carthaca:
Thanks for bringing this up :)

guigui89:
Thank you for the responses. When is going to be a new release date for fog-openstack?

PR #239, opened 2016-11-25
add size to compute image model

scambra:
Add size attribute reading OS-EXT-IMG-SIZE:size


seanhandley:
Thanks @scambra. Confirmed this is in the API reference: http://developer.openstack.org/api-ref/compute/?expanded=show-image-details-detail#show-image-details

PR #218, opened 2016-10-13
Manila actions

Carthaca:
clean replacement of #212 after renaming to "shared file system"

@gildub fyi

PR #242, opened 2016-11-28
introduce microversion

Carthaca:
starting with manila

split off https://github.com/fog/fog-openstack/pull/218

### DIFF
--- a/lib/fog/openstack.rb
+++ b/lib/fog/openstack.rb
@@ -537,63 +537,25 @@ module Fog
     end
 
     def self.get_supported_version(supported_versions, uri, auth_token, connection_options = {})
-      connection = Fog::Core::Connection.new("#{uri.scheme}://#{uri.host}:#{uri.port}", false, connection_options)
-      response = connection.request({
-                                        :expects => [200, 204, 300],
-                                        :headers => {'Content-Type' => 'application/json',
-                                                     'Accept' => 'application/json',
-                                                     'X-Auth-Token' => auth_token},
-                                        :method => 'GET'
-                                    })
-
-      body = Fog::JSON.decode(response.body)
-      version = nil
-      unless body['versions'].empty?
-        versions = body['versions'].kind_of?(Array) ? body['versions'] : body['versions']['values']
-        supported_version = versions.find do |x|
-          x["id"].match(supported_versions) &&
-            (x["status"] == "CURRENT" || x["status"] == "SUPPORTED" || x["status"] == "stable")
-        end
-        version = supported_version["id"] if supported_version
-      end
-      if version.nil?
-        raise Fog::OpenStack::Errors::ServiceUnavailable.new(
-                  "OpenStack service only supports API versions #{supported_versions.inspect}")
-      end
+      supported_version = get_version(supported_versions, uri, auth_token, connection_options)
+      version = supported_version['id'] if supported_version
+      version_raise(supported_versions) if version.nil?
 
       version
     end
 
     def self.get_supported_version_path(supported_versions, uri, auth_token, connection_options = {})
-      # Find a version in the path (e.g. the v1 in /xyz/v1/tenantid/abc) and get the path up until that version (e.g. /xyz))
-      path_components = uri.path.split '/'
-      version_component_index = path_components.index{|comp| comp.match(/v[0-9].?[0-9]?/) }
-      versionless_path = (path_components.take(version_component_index).join '/' if version_component_index) || uri.path
-      connection = Fog::Core::Connection.new("#{uri.scheme}://#{uri.host}:#{uri.port}#{versionless_path}", false, connection_options)
-      response = connection.request({
-                                        :expects => [200, 204, 300],
-                                        :headers => {'Content-Type' => 'application/json',
-                                                     'Accept' => 'application/json',
-                                                     'X-Auth-Token' => auth_token},
-                                        :method => 'GET'
-                                    })
-
-      body = Fog::JSON.decode(response.body)
-      path = nil
-      unless body['versions'].empty?
-        versions = body['versions'].kind_of?(Array) ? body['versions'] : body['versions']['values']
-        supported_version = versions.find do |x|
-          x["id"].match(supported_versions) &&
-              (x["status"] == "CURRENT" || x["status"] == "SUPPORTED")
-        end
-        path = URI.parse(supported_version['links'].first['href']).path if supported_version
-      end
-      if path.nil?
-        raise Fog::OpenStack::Errors::ServiceUnavailable.new(
-                  "OpenStack service only supports API versions #{supported_versions.inspect}")
-      end
+      supported_version = get_version(supported_versions, uri, auth_token, connection_options)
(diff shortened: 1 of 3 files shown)


gildub:
@Carthaca ,

First off, thank you for creating a separate PR.
The re-factoring part is good as I think it's really needed.

The user need to be able to pass a version number:
* My understanding of the micro-versions is to be able for users to have stability by sticking to a specific version and at same time, it frees developers by allowing them to push new features by bumping up (micro)versions.

I also think that most of the microversion logic can be in the upper class and reduce impact on subclasses. Effectively, all Openstack projects using micro version are exactly the same, the only difference is in the micro version header (`X-Openstack-API-Version` and/or `X-Openstack-<SERVICE>-API-Version`).

What do you think?

Carthaca:
@gildub, I agree, I will move as much as possible to `Fog::OpenStack::Core`.

`set_api_path` is another candidate for such a move - I will tackle that in a separate PR later, I hope.

To the ability of passing a microversion number: I understand that some might want that.
I guess I can come up with a solution that supports that, too - though I don't see the necessity for my use-case.

Carthaca:
`openstack_shared_file_system_microversion` allows to wish for a specific microversion - it will produce a warning if the backend or the supported microversion is lower and continue to use the maximum supported microversion instead

Carthaca:
@gildub can you have a look please?

seanhandley:
Glad to see this is progressing! This is an important feature and it's important we get it added.

I know it's a busy month @gildub but your input would be awesome 👍

gildub:
Review in progress, thanks for your patience :)

lib/fog/openstack/core.rb
```diff
@@ -119,6 +116,56 @@ def set_api_path
...
+        # choose minimum out of set and wished version
+        if @fixed_microversion && microversion_newer_than?(@fixed_microversion)
+          @microversion = @fixed_microversion
+        elsif @fixed_microversion && @microversion != @fixed_microversion
+          Fog::Logger.warning("Microversion #{@fixed_microversion} not supported")
+        end
+      end
+
+      def microversion_newer_than?(version)
+        Gem::Version.new(version) < Gem::Version.new(@microversion)
```
gildub: Also using Gem `version` this way works, I don't think that's the standard approach.
Effectively it escapes Gem dependency (handled through gem specs). That would also need a `require` statement.

lib/fog/openstack.rb
```diff
@@ -614,6 +576,45 @@ def self.map_endpoint_type type
...
+      # order is important, preffered status should be first
+      %w(CURRENT stable SUPPORTED DEPRECATED).each do |status|
+        version = versions.find { |x| x['id'].match(supported_versions) && (x['status'] == status) }
+        break if version
+      end
+
+      version
+    end
+
+    def self.get_version_values(versions)
```
gildub: Why such method? It's called only once and it has only one line of code.

lib/fog/openstack/core.rb
```diff
@@ -119,6 +116,56 @@ def set_api_path
         # if the service supports multiple versions, do the selection here
       end
 
+      def set_microversion
+        # both need to be set in service's initialize for microversions to work
+        return if !@supported_microversion || !@supported_versions
```
gildub: What about doing that test before, at the calling time (Line 91)? 
Because the job of set_microversion is not to test if it needs to be done, I think it's a decoupling thing.

lib/fog/openstack/core.rb
```diff
@@ -119,6 +116,56 @@ def set_api_path
...
+
+        @microversion_key          ||= 'Openstack-API-Version'.freeze
+        @microversion_service_type ||= @openstack_service_type.first
```
gildub: Sorry I'm not seeing why we need the service type here?

lib/fog/shared_file_system/openstack.rb
```diff
@@ -250,28 +251,37 @@ def self.not_found_class
...
+          initialize_identity options
 
-          @connection_options               = options[:connection_options] || {}
+          @openstack_service_type  = options[:openstack_service_type] || ['sharev2']
+          @openstack_service_name  = options[:openstack_service_name]
+          @connection_options      = options[:connection_options] || {}
 
           authenticate
           set_api_path
+          set_microversion
```
gildub: This is not necessary since the base class is doing it anyway.

gildub:
First off, sorry for not being available sooner to provide feedback.

Besides in-line comments, the main thing is that I don't see one can specify a required version which I  believe is a common use case for the users as it's a main feature of micro versions.
Any chance a user to specify a (micro)version by number(V2.15) or by state (SUPPORTED, LATEST, CURRENT)? Maybe using the params.

lib/fog/openstack/core.rb
gildub: Ok, I just ignore my previous comment, I realized that Version is part of stdlib with rubygems anyway.

lib/fog/openstack.rb
Carthaca: correct, no need for that anymore - in the middle of refactoring it was used two times ;)

lib/fog/openstack/core.rb
Carthaca: good point

lib/fog/openstack/core.rb
Carthaca: at line https://github.com/sapcc/fog-openstack/blob/bce4837ac29502affd4d5ea6c5be00cb2378756e/lib/fog/openstack/core.rb#L154

for how [compute](http://developer.openstack.org/api-guide/compute/microversions.html#client-interaction)/[identity](https://specs.openstack.org/openstack/api-wg/guidelines/microversion_specification.html#client-interaction) use the microversion with 'Openstack-API-Version' Header in contrast to [shared file system](http://docs.openstack.org/developer/manila/devref/api_microversion_dev.html), where we don't have that (yet).
And I need to be able to specify it for each service, because it can differ from the service_type (f.i. 'compute' vs. 'computev2')

Carthaca:
Good remark to the calling of set_microversion, I added it with the conditional to be always perfomed on authentication - else the reaction after token expiration would be wrong as it looks now.

And to the specification: you can set `openstack_shared_file_system_microversion` to set the microversion for manila by number. It will have to be introduced by each service, because it is service independent (i.e. if I set my credentials and want to use multiple services that have microversioning I need to be able to specify multiple microversions in parallel).

If you don't specify it, then we will always take the minimum out of the reported microversion by the backend and the microversion specified in code as supported (see https://github.com/sapcc/fog-openstack/blob/bce4837ac29502affd4d5ea6c5be00cb2378756e/lib/fog/openstack/core.rb#L133-L136)

Hope this is clear enough :)

Carthaca:
I've put the fix for the ruby 1.9 test into https://github.com/fog/fog-openstack/pull/254

seanhandley:
I guess this needs a rebase with master now #254 is merged?

Carthaca:
done

seanhandley:
I'm happy with this if you are @gildub

seanhandley:
We should also merge #218 when this is merged.

PR #246, opened 2016-12-06
Handle NotFound in images v2 model's `get`

beyhan:
This is consistent with the behavior of `get` in other
 models, e.g. servers, volumes.

Signed-off-by: Beyhan Veli <beyhan.veli@sap.com>

### DIFF
--- a/lib/fog/image/openstack/v2/models/images.rb
+++ b/lib/fog/image/openstack/v2/models/images.rb
@@ -18,6 +18,8 @@ module Fog
 
           def find_by_id(id)
             new(service.get_image_by_id(id).body)
+          rescue Fog::Image::OpenStack::NotFound
+            nil
           end
 
           alias get find_by_id
--- a/spec/fixtures/openstack/image_v2/images_v2_find_by_id.yml
+++ b/spec/fixtures/openstack/image_v2/images_v2_find_by_id.yml
@@ -1,0 +1,81 @@
+---
+http_interactions:
+- request:
+    method: get
+    uri: http://devstack.openstack.stack:9292/v2/images/fe05659e-d433-4e09-aa78-19e0b7f5e497
+    body:
+      encoding: US-ASCII
+      string: ''
+    headers:
+      User-Agent:
+      - fog-core/1.42.0
+      Content-Type:
+      - application/json
+      Accept:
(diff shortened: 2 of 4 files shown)


voelzmo:
cc @dhague and @Ladas

dhague:
It's not consistent with Identity, Monitoring and very many unit tests though.
As you say, there are also a lot of places where we swallow the *::OpenStack::NotFound exceptions.
Personally, I don't see the point in defining such an exception if we only ever swallow it - so the question becomes: in general do we want to have NotFound exceptions (in which case we should raise them all the way to user code), or should we just return nil instead (which is the practical effect to user code of swallowing them like we do in several places right now). 
One thing is for sure: I think we have a contradiction in having several unit tests which expect a NotFound exception (raised in the Mock request code), but we just return nil in the real-life invocation of that request.

voelzmo:
Hey @dhague thanks for the comments.

The concerns you raise about consistency across all resources (including things like identity and monitoring), the tests, and mock classes are very valid and definitely need discussion. Whatever the outcome of that discussion will be, it will definitely be a larger operation to harmonize all the code and will most likely require a major release because of breaking API changes.

In this PR, however, we wanted to have image v1 consistent with image v2 (and with many 'core' OpenStack entities, such as servers, volumes, snapshots, networks, ports, etc). I assume that we're not the only ones having tooling that deals with OpenStack installations that might either have image v1 or v2. Right now, we have a separate codepath for image deletion if it is v2 – that seems like the worst of options for now.

We're happy to join a larger discussion. Can we have that after merging this tiny thing that just has image v1 and image v2 act the same?

dhague:
@voelzmo thanks for explaining further. If you could take a small amount of time to add specific tests for v1 & v2 to show that v1 & v2 would have the same behavior then I would have no problem merging this PR.

beyhan:
@dhague We added the requested tests. Let us know if that works for you.

dhague:
Sorry to nitpick, but please add the same tests to v2 as well - that will make it crystal clear that the expected behavior is the same in each case.

beyhan:
@dhague added tests for images v2.

dhague:
Merged - thanks!

PR #249, opened 2016-12-15
removed extraneous tenant_id from quota_set in update_quota

sseago:
In both the Compute and Volume update_quota actions, update_quota
was adding tenant_id to the quota_set hash. It doesn't belong there.
tenant_id is already in the URL and it doesn't go into the quota
hash. For the Compute operation, adding it to quota_set causes it
to fail. For Volume, it doesn't fail, but it still doesn't belong there.

### DIFF
--- a/lib/fog/compute/openstack/requests/update_quota.rb
+++ b/lib/fog/compute/openstack/requests/update_quota.rb
@@ -3,7 +3,6 @@ module Fog
     class OpenStack
       class Real
         def update_quota(tenant_id, options = {})
-          options['tenant_id'] = tenant_id
           request(
             :body    => Fog::JSON.encode('quota_set' => options),
             :expects => 200,
--- a/lib/fog/volume/openstack/requests/update_quota.rb
+++ b/lib/fog/volume/openstack/requests/update_quota.rb
@@ -3,7 +3,6 @@ module Fog
     class OpenStack
       module Real
         def update_quota(tenant_id, options = {})
-          options['tenant_id'] = tenant_id
           request(
             :body    => Fog::JSON.encode('quota_set' => options),
             :expects => 200,


seanhandley:
Thanks @sseago. Has this been tested on non Redhat versions of OpenStack?

I'm using Keystone's quota commands in production and they've worked on Juno, Kilo, and Mitaka (I'm referring to this here: https://bugzilla.redhat.com/show_bug.cgi?id=1390147)

seanhandley:
The documentation for Nova definitely confirms that the `tenant_id` param should be in the URL and not the request body. I'm just confused why it works for some OpenStacks and not others...

sseago:
@seanhandley I'm not sure exactly which versions this bug has been seen with. I was able to reproduce the bug using OSP8 (Liberty). Regarding why it appears to work in some cases without this fix, note that the code wasn't simply placing tenant_id in the request body -- it was placing it within the quota_set hash in the request body. This is wrong even if tenant_id belongs in the body (but, as you note, it belongs in the URL). Since it was also in the URL, though, the only issue here is whether the tenant_id added to the quota_set triggers an error or is ignored. At least on OSP8, I was seeing that adding tenant_id to the quota_set for nova calls (*compute* update_quota in fog-openstack) triggered an "Additional properties are not allowed" error, while adding tenant_id to the quota_set for cinder calls (*volume* update_quota in fog-openstack) does not trigger an error -- the operation succeeds, presumably ignoring that extra tenant_id param entirely.

It could be that the different behavior across openstack versions is related to validation of unexpected parameters (are they ignored or do they result in an error).

seanhandley:
Well I'm confident it matches the API reference so I'm happy to merge. Thanks for this!

PR #251, opened 2016-12-16
allow subnet id parameter on floating ip allocation

Carthaca:


seanhandley:
Thanks @Carthaca. I can't see any examples for sending a `subnet_id` to the API when creating a floating IP. I can see it mentioned in the documentation briefly here: http://developer.openstack.org/api-ref/networking/v2/?expanded=create-floating-ip-detail

If you could point me towards some examples or something then I'm happy to review.

Carthaca:
I'm a little bit surprised, that the api-ref is missing that - probably I will add it there, too :)

I see that this parameter can be used in the official python command line client: https://github.com/openstack/python-openstackclient/blob/47716d1ad32b7c879f6b5736687be1553f6fbfb6/openstackclient/network/v2/floating_ip.py#L54

seanhandley:
That works! Thanks 👍

Carthaca:
official api-ref fix will come with https://review.openstack.org/#/c/413346/

PR #254, opened 2016-12-27
pin down nokogiri gem for ruby 1.9

Carthaca:
fix failing test for ruby 1.9: nokogiri 1.7.0 was released yesterday and dropped ruby < 2 support (https://github.com/sparklemotion/nokogiri/blob/master/CHANGELOG.md#backwards-incompatibilities)


seanhandley:
Thanks @Carthaca - sorry this took so long to merge.

PR #262, opened 2017-01-23
Added mocks for Fog::Storage::OpenStack

krzyczak:
1. The class `Mock` was added in this module where needed
2. Added `fail "Mock Not Implemented (#method_name) in: #{__FILE__}:#{__LINE__}"` where mock was not implemented

Code is mostly coppied from `Fog::Storage::Rackspace::Mock`.

### DIFF
--- a/lib/fog/storage/openstack/requests/delete_container.rb
+++ b/lib/fog/storage/openstack/requests/delete_container.rb
@@ -1,6 +1,12 @@
 module Fog
   module Storage
     class OpenStack
+      class Mock
+        def delete_container(name)
+          fail "Mock Not Implemented (#delete_container) in: #{__FILE__}:#{__LINE__}"
+        end
+      end
+
       class Real
         # Delete an existing container
         #
--- a/lib/fog/storage/openstack.rb
+++ b/lib/fog/storage/openstack.rb
@@ -44,6 +44,201 @@ module Fog
       request :public_url
 
       class Mock
+        class MockContainer
+          attr_reader :objects, :meta, :service
+
+          # Create a new container. Generally, you should call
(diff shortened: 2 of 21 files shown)


lib/fog/storage/openstack/requests/delete_container.rb
```diff
@@ -1,6 +1,12 @@
 module Fog
   module Storage
     class OpenStack
+      class Mock
+        def delete_container(name)
+          fail "Mock Not Implemented (#delete_container) in: #{__FILE__}:#{__LINE__}"
```
plribeiro3000: I know you did a copy and paste but you dont need to define this. Once It is not implemented fog will throw the error automatically. =)

So you can skip this change.

krzyczak: Actually, I added this myself to see where was the exception from. Fog only showed sometinhg like `Not yet implemented` and I wanted to get the exact location of the not implemented thing.

plribeiro3000: Gotcha. Maybe you might want to add this to fog-core instead so all providers will have this feature?

Also, does this work with all active maintained ruby versions?

ShamoX:
@plribeiro3000 & @krzyczak & @timuralp : Where does this PR stands ?

plribeiro3000:
@ShamoX As a first look it needs rebase.

considering behaviour we would need someone with access to `openstack` to test it as i do not have it.

timuralp:
The mocking bits in the PR seem fine to me. Adding the additional auth parameters for keystone is probably also fine, but we don't include the tests for it here? Unfortunately, I don't think anyone is actively looking into this, though.

PR #267, opened 2017-02-01
blank is not a standard ruby method, so replacing it with equivalent …

kowsalyapalaniappan:
blank is not a standard ruby method, so replacing it with equivalent conditional statements


Carthaca:
thanks

PR #268, opened 2017-02-01
Allow to connect to identity_v3 endpoint even if it's not in the catalog

achevalet:
It may happen that identityv3 endpoint is not present in the catalog (but only v2 for local users) and public users have to connect to v3 endpoint, which is "often" the auth url - so try it

### DIFF
--- a/lib/fog/openstack.rb
+++ b/lib/fog/openstack.rb
@@ -259,7 +259,7 @@ module Fog
 
     # Keystone Style Auth
     def self.authenticate_v3(options, connection_options = {})
-      uri = options[:openstack_auth_uri]
+      uri                   = options[:openstack_auth_uri]
       project_name          = options[:openstack_project_name]
       service_type          = options[:openstack_service_type]
       service_name          = options[:openstack_service_name]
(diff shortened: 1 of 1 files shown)


aufi:
If I understand this PR correctly, the side effect is that endpoint list returned to client contains identity endpoint _only_.

Couldn't we a have a switch to enable/disable this "fallback" for Openstack deployments without proper identity v3 endpoint?

gildub:
@aufi, 

Good question! To be confirmed and fixed if needed, the default behaviour should be to use v3, even if endpoint is missing, unless v2.0 is specified, because v3 is the default as v2.0 has been deprecated.

So from that we shouldn't have to use a switch, provided the whole catalog is returned no matter what.

gildub:
@achevalet,

Could you please look at it?

achevalet:
If the auth url contains "v3", fog gets identityv3 url from the catalog and uses it by default. 
If v3 endpoint is not listed in the catalog, it uses the auth url (removing /auth/token, ex: https://openstack:5000/v3) as a last try before failing. It does not fallback on v2. So from what I can see, there is no need of a switch.

achevalet:
ps: it's the behavior of most of openstack SDKs

gildub:
@achevalet, agreed on the v3 being the default even if service is missing form catalog, the concern is about your patch and the catalog being returned containing *only* the identity service in the list. I haven't had a chance to test it.

achevalet:
If there is only 'identity' endpoint it will not be used. 'identity' is
used only if the auth_url is v2. My patch didn't change that behavior.

2017-05-18 2:46 GMT+02:00 Gilles Dubreuil <notifications@github.com>:

gildub:
@achevalet, the concern was about the content of the catalog, meanwhile I've double checked and that's fine, therefore never mind. Thanks 

@aufi, I ran the test against OSP9 which has no identityv3 endpoint set by default and the catalog is fulfilled. What's always a bit confusing is the bootstrap process for initial authentication.

PR #281, opened 2017-03-16
Neutron Request update_router: Clear gateway

gildub:
The options allows to reset (clear) the gateway details.

https://bugzilla.redhat.com/show_bug.cgi?id=1426268

### DIFF
--- a/lib/fog/network/openstack/requests/update_router.rb
+++ b/lib/fog/network/openstack/requests/update_router.rb
@@ -15,7 +15,6 @@ module Fog
             data['router'][key] = options[key] if options[key]
           end
 
-          # remove this in a future
           egi = options[:external_gateway_info]
           if egi
             if egi.kind_of?(Fog::Network::OpenStack::Network)
@@ -27,6 +26,9 @@ module Fog
             else
               raise ArgumentError, 'Invalid external_gateway_info attribute'
             end
+          else
+            # Allow to clear gateway
+            data['router'][:external_gateway_info] = nil
           end
 
           request(


gildub:
@Carthaca ,

This is the right way to clear the gateway information not an empty dict. 

Ruby `nil` is translated into JSON `null` and that's what must be provided to clear the gateway information. A empty Dict is wrong and didn't work for me anyway (but even if it does it's a side effect).

For more information please check [1].

Thanks

```ruby
 "routers": [
        {
            "admin_state_up": true,
            "availability_zone_hints": [],
            "availability_zones": [
                "nova"
            ],
            "description": "",
            "distributed": false,
            "external_gateway_info": null,
            "ha": false,
            "id": "13e4d26f-3660-43e8-9794-3545f2a67081",
            "name": "therouter1234",
            "routes": [],
            "status": "ACTIVE",
            "tenant_id": "48985e6b8da145699d411f12a3459fca"
        }]
```

Assuming latter router is define and running this CURL script

```shell
$ cat > update-router.sh <<EOF
OS_TOKEN=92fb355227b44a0a85efad48bf4a097a
OS_HOST=192.0.2.21
ROUTER=13e4d26f-3660-43e8-9794-3545f2a67081

# Update a router
curl -s \
  -X PUT \
  -H "X-Auth-Token: $OS_TOKEN" \
  -H "Content-Type: application/json" \
  -d '{ "router": { "name": "router12345", "admin_state_up":true, "external_gateway_info": null } }' \
http://${OS_HOST}:9696/v2.0/routers/${ROUTER} | python -mjson.tool

{
    "router": {
        "admin_state_up": true,
        "availability_zone_hints": [],
        "availability_zones": [
            "nova"
        ],
        "description": "",
        "distributed": false,
        "external_gateway_info": null,
        "ha": false,
        "id": "13e4d26f-3660-43e8-9794-3545f2a67081",
        "name": "router12345",
        "routes": [],
        "status": "ACTIVE",
        "tenant_id": "48985e6b8da145699d411f12a3459fca"
    }
}
```

[1] https://developer.openstack.org/api-ref/networking/v2/?expanded=update-router-detail

Carthaca:
What neutron are you working on? We are on mitaka and it works with the empty hash.
I'm happy to change it to do it the right way and not use the accidental way.

But, please concentrate on the second part of my statement - a simple router name update would clear the gateway.

gildub:
@Carthaca,

Oh, I see what you mean, my bad,
I guess I was *too* concentrated on clearing the gateway data!
Also after rechecking, it *does* work with an empty list, my test was wrong too :(
So we can leave it for now. I'll run more tests tomorrow.

Cheers

gildub:
This is effectively not necessary, an empty list `:external_gateway_info=> {}` works fine. Thanks

PR #283, opened 2017-04-17
List metric resources by network interface type

rwsu:
Add an additional request to list only network interface resources.

### DIFF
--- a/lib/fog/metric/openstack/requests/list_resources.rb
+++ b/lib/fog/metric/openstack/requests/list_resources.rb
@@ -6,7 +6,7 @@ module Fog
           request(
             :expects => 200,
             :method  => 'GET',
-            :path    => "resource/#{type}",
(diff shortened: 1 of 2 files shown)


Carthaca:
I don't think we should introduce custom metric resources here. You should be able to get the same result by filtering the generic endpoint for the resource type, from how I read this: https://docs.openstack.org/developer/gnocchi/rest.html#resources

So something like `list_resources(:type => 'instance_network_interface')` should work, I hope

If this doesn't work, I suggest introducing the type as parameter to set the path to have this more flexible

rwsu:
@Carthaca Unfortunately, list_resources(:type => 'instance_network_interface') does not work. I will update the PR to add the type as a parameter instead of introducing a new request. Thanks for clearing it up. After looking at the existing examples, I wasn't sure whether the preferred approach is to add a new parameter or new request.

Carthaca:
Thanks. I just added escaping for the type

Issue #278, opened 2017-02-27
Add support for Barbican Secret ACLs


ScottMaclure:
https://docs.openstack.org/developer/barbican/api/reference/acls.html

My Use Cases:

* Allows for project-access = false
* Can add specific Users to a Secret ACL

Would this go into KeyManager?

https://github.com/fog/fog-openstack/blob/master/lib/fog/key_manager/openstack.rb

Carthaca:
Hi @ScottMaclure, yes KeyManager is the right place.

Feel free to implement and send a Pull-request :)

PR #286, opened 2017-04-25
Add support for Barbican Secret ACLs 

kowsalyapalaniappan:
Added support to update, delete and get acl details of secrets and containers. issue https://github.com/fog/fog-openstack/issues/278


lib/fog/key_manager/openstack.rb
```diff
@@ -41,6 +42,12 @@ class OpenStack < Fog::Service
       request :list_containers
       request :delete_container
 
+      #acl
+      request :get_acl
```
Carthaca: please list the available requests (e.g. `get_secret_acl`) to make them discoverable, not the file names

lib/fog/key_manager/openstack/models/acl.rb
```diff
@@ -0,0 +1,24 @@
+
+require 'fog/openstack/models/model'
+require 'uri'
```
Carthaca: unnecessary require

lib/fog/key_manager/openstack/models/acl.rb
```diff
@@ -0,0 +1,24 @@
...
+
+module Fog
+  module KeyManager
+    class OpenStack
+
+      class Acl < Fog::OpenStack::Model
```
Carthaca: I suggest using upper case `ACL` here

lib/fog/key_manager/openstack.rb
kowsalyapalaniappan: made this change in latest commit.

lib/fog/key_manager/openstack/requests/get_acl.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      end
+
+      class Mock
+
+        def get_secret_acl(_uuid)
+          response = Excon::Response.new
+          response.status = 200
+          response.body = {
+              "data" => {
```
Carthaca: Thanks for adding some mocks :) Would you please format them a bit more? And I believe the keys `data` and `body` are wrong here, i.e. not needed to wrap the response's body

kowsalyapalaniappan: i splited the files and updated the mocks.

Carthaca:
thanks, I just added the mock class and fixed some style

kowsalyapalaniappan:
Thank you

PR #288, opened 2017-05-12
Extract custom properties when searching for images

zeroDivisible:
This patch fixes the logic when finding an image by ID, using the V1
version of Glance API. Response from Glance when getting details of
an image contains headers prefixed by one of two strings:

  - X-Image-Meta-
  - X-Image-Meta-Property

The second one is used for enumerating all the custom properties
associated with given image and the current version of the code wasn't
correctly handling them.

The source of the bug was that because `X-Image-Meta-` was used as the
prefix when selecting headers, all the custom properties were selected
as well - and then incorrectly treated like normal properties for that
image.

### DIFF
--- a/lib/fog/image/openstack/v1/models/images.rb
+++ b/lib/fog/image/openstack/v1/models/images.rb
@@ -23,10 +23,21 @@ module Fog
           end
 
           def find_by_id(id)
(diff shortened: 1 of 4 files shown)


gildub:
LGTM

Carthaca:
👍

PR #290, opened 2017-05-16
Handle nil values of data[:server_groups] in mock delete_server

zeroDivisible:
This patch modifies the logic in the mock implementation of deleting a server.
The original code has nil checks for making sure that we will only remove a
server from a group if that group exists - but it does not check whether the
whole collection of groups exist.

This can lead to a situation that if the user data when mocking doesn't contain
the key `data[:server_groups]`, we'd be hitting the nil pointer exception on the
calls to

  data[:server_groups].find

This only affects the mock implementation.


Carthaca:
thanks

zeroDivisible:
No worries, happy to help!

PR #293, opened 2017-05-18
add snapshot_action to cinder

Carthaca:
same like on volumes one can perform certain actions on cinder snapshots

### DIFF
--- a/lib/fog/volume/openstack/requests/snapshot_action.rb
+++ b/lib/fog/volume/openstack/requests/snapshot_action.rb
@@ -1,0 +1,16 @@
+module Fog
+  module Volume
+    class OpenStack
+      module Real
+        def snapshot_action(id, data)
+          request(
+            :body    => Fog::JSON.encode(data),
+            :expects => [200, 202],
+            :method  => 'POST',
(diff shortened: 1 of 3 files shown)


gildub:
Looks good. 
What about the Mock? Look, I'm not a big fan as I'd like to see them being separated from the real code meanwhile until that happens, don't we we have to stay consistent ?

gildub:
Actually that reminded me to reactivate https://github.com/fog/fog-openstack/pull/223

Carthaca:
It's the minimal method on which you could build something on top like [extend volume](https://github.com/fog/fog-openstack/blob/master/lib/fog/volume/openstack/requests/extend_volume.rb) (which should be refactored to actually use the action method btw)

I fear the `snapshot_action` method itself would be hard to mock, because it is too flexible.
Same like with the [action](https://github.com/fog/fog-openstack/blob/master/lib/fog/volume/openstack/requests/action.rb) method, I would just like to skip this

gildub:
@Carthaca,

Sorry the head in the handle bar, UI stuff is *way* too consuming, anyway I understand the mock would be mockery anyway ;)

PR #301, opened 2017-06-07
[manila] security service

Carthaca:


lib/fog/shared_file_system/openstack/requests/share_network_action.rb
```diff
@@ -0,0 +1,16 @@
+module Fog
+  module SharedFileSystem
+    class OpenStack
+      class Real
+        def share_network_action(id, options = {}, expects_status = 202)
```
gildub: Why "202" as default, I can't see that in the API ref.

gildub:
I understand the DRY intent between the "Add security service to share network" and "Remove security service from share network" requests, but I'm not comfortable with it maybe because it's using request method which doesn't exist as such instead of using an internal method. Maybe it's just me, anyone?

lib/fog/shared_file_system/openstack/requests/share_network_action.rb
Carthaca: That 202 is just a copy&paste - will change to 200

gildub:
@Carthaca, would you like to update it?

gildub:
Since nobody has responded and challenged this, then let's move forward.

Carthaca:
Thanks and sorry for my silence - I can't fully contribute to this project any more

gildub:
@Carthaca, no worries, we're going to miss you, good luck then!

PR #303, opened 2017-06-23
Allow more than 100 results post-Liberty.

seanhandley:
The Liberty release of Ceilometer introduced the idea of a limit variable to tell the API how many results you want. By default this is set to 100 in the server config. The onus is on the client to select how many results it wants.

I propose we default to a very high number as this will provide minimal disturbance to users.

Read this: https://specs.openstack.org/openstack/ceilometer-specs/specs/liberty/mandatory-limit.html

and this: https://docs.openstack.org/developer/ceilometer/webapi/v2.html#get--v2-meters-(meter_name)

### DIFF
--- a/lib/fog/metering/openstack/requests/get_samples.rb
+++ b/lib/fog/metering/openstack/requests/get_samples.rb
@@ -2,7 +2,7 @@ module Fog
   module Metering
     class OpenStack
       class Real
-        def get_samples(meter_id, options = [])
+        def get_samples(meter_id, options = [], limit = 10000000)
           data = {
             'q' => []
           }
@@ -15,6 +15,7 @@ module Fog
             end
 
             data['q'] << filter unless filter.empty?
+            data['limit'] = limit
           end
 
           request(


lib/fog/metering/openstack/requests/get_samples.rb
```diff
@@ -2,7 +2,7 @@ module Fog
   module Metering
     class OpenStack
       class Real
-        def get_samples(meter_id, options = [])
+        def get_samples(meter_id, options = [], limit = 10000000)
```
gildub: Wouldn't it be possible to avoid changing the interface using `options = [:limit => 10000000]` instead?

gildub:
@seanhandley, Sounds good. Just wondering about having the parameter inside options (see inline comment) ?

lib/fog/metering/openstack/requests/get_samples.rb
seanhandley: No - it sets the `data` hash with `q` and then JSON encodes it as the HTTP post body.

seanhandley: The relevant API doc is referenced in the PR body.

seanhandley: Since the `option` hash in our method becomes the `q` JSON hash which has a specific schema to query ceilometer. The `limit` param is entirely separate from that.

So there's likely client code out there that looks like this:

```ruby
timestamp_format = "%Y-%m-%dT%H:%M:%S"
options = [{'field' => 'timestamp', 'op' => 'ge', 'value' => from.utc.strftime(timestamp_format)},
                   {'field' => 'timestamp', 'op' => 'lt', 'value' => to.utc.strftime(timestamp_format)}]
Fog::Metering.new(params).metering.get_samples(measurement, options)
```

That array becomes the `q` contents in the post body. To stay backwards compatible I think the results limit has to be a separate optional arg.

gildub: That makes sense, thanks for the explanation!

PR #304, opened 2017-06-23
Remove ".json" extension from Nova API requests

mansam:
The routes in the Nova API that support ".json" in the path are undocumented and are being removed as a consequence of refactoring the Nova API to remove stevedore.[1] As a consequence, certain Fog requests fail for new versions of Nova. This PR removes ".json" from requests to the Nova API.

I don't have an adequate way to fully test the consequences of this on every supported version of Nova.

@Ladas

[1] http://lists.openstack.org/pipermail/openstack-dev/2017-March/114736.html

### DIFF
--- a/lib/fog/compute/openstack/requests/add_flavor_access.rb
+++ b/lib/fog/compute/openstack/requests/add_flavor_access.rb
@@ -9,7 +9,7 @@ module Fog
                                          }),
             :expects => [200, 203],
             :method  => 'POST',
-            :path    => "flavors/#{flavor_ref}/action.json"
+            :path    => "flavors/#{flavor_ref}/action"
           )
         end
       end
--- a/lib/fog/compute/openstack/requests/allocate_address.rb
+++ b/lib/fog/compute/openstack/requests/allocate_address.rb
@@ -7,7 +7,7 @@ module Fog
             :body    => Fog::JSON.encode('pool' => pool),
             :expects => [200, 202],
             :method  => 'POST',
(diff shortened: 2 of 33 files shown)


gildub:
@mansam,

👍 LGTM

I believe there is no real need for documenting the JSON extension because it's an HTTP convention whether to let consumers choose the format using either the HTTP Accept header or by adding to the URL the `.json` extension. It should only be a matter of preference.

`fog-openstack` is effectively using a mixture of both which is not consistent and actually the same clean up should be applied across all services.

Anyway, "Accept" header is always added [1] therefore all `.json` extension can be removed safely from any path.

Because of the level of impact, this need to be confirmed by someone else before merging. 

Thanks

[1] https://github.com/fog/fog-openstack/blob/master/lib/fog/openstack/core.rb#L160

Carthaca:
LGTM, I agree  - should effectively be done across all services

mansam:
@gildub  If adding a file extension to a path to request a particular content type is understood widely to be an http convention, then I think the change to the Openstack API to remove it should be considered as breaking virtually-documented behavior due to the convention, and we should be much more annoyed ;). I don't think the convention is actually that common due to the specific support it requires. For what it's worth, the Neutron API reference explicitly documents this feature.

gildub:
@mansam, I understand and "How a practice can be a convention if it's not used any more?".

aufi:
@mansam @gildub Manually applied and tested - works great! Looking forward to have released a new version of ```fog-openstack``` gem.

PR #309, opened 2017-07-17
Swift: Fix existing requests storage tests to work live.

timuralp:
Fixes for the existing storage tests to pass against a Swift cluster. Tested all of the methods, except for the temp URL, as I did not have it configured on the cluster. Will address the tempurl failures in a subsequent patch.

### DIFF
--- a/test/requests/storage/container_tests.rb
+++ b/test/requests/storage/container_tests.rb
@@ -1,9 +1,20 @@
 require "test_helper"
 
 describe "Fog::Storage[:openstack] | container requests" do
+  def cleanup_container
+    return if Fog.mocking?
+    if @storage.head_container(@container_name)
+      @storage.delete_container(@container_name)
+    end
+  rescue Fog::Storage::OpenStack::NotFound
+  end
+
   before do
     @storage = Fog::Storage[:openstack]
     @container_format = [String]
+    @container_name = 'fogcontainertests'
+
+    cleanup_container
 
     @containers_format = [{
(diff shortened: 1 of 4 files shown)


timuralp:
@seanhandley I mutated the existing, non-passing storage requests tests to pass against a Swift cluster. Some of the tests do create and delete containers and objects to test certain behavior (e.g. container and object must exist to test the result of HEAD). Prior iteration assumed in-order execution and state carrying through between them. I made the tests be allowed to be executed in random order (they will fail if run in parallel because they all use the same container/object names), but it's a step forward. Let me know if there is a cleaner paradigm to follow when writing these in the future.

timuralp:
@seanhandley @geemus I think since I'm the only person working on the fog OpenStack Swift support, it might make sense to let me be able to commit changes to this gem? Since fixing the live tests, I'd like to go back and see how difficult it would be to add mocks -- there is already a PR for that from someone. And then start looking into what it would take to bring fog Swift support to parity with the python Swift client.

geemus:
@timuralp makes sense to me. I just sent you an invite for the fog-openstack team, which I think should get you what you need. Thanks for your ongoing help!

timuralp:
@geemus thank you! This will make it a bit faster to iterate on improvements (although, I can't claim that this will be my full time endeavor).

timuralp:
Verified locally against Swift all-in-one. Merging this PR to get live tests to work.

geemus:
Thanks!

PR #314, opened 2017-09-23
[WIP] Fix HTTP body params on GET queries

aufi:
There are some queries which use HTTP GET method, but pass parameters
in request body. This approach is not correct and might raise errors on
smarter firewalls.

Updating code moving parameters from body to query for Metering and Event
services.

### DIFF
--- a/lib/fog/event/openstack/requests/list_events.rb
+++ b/lib/fog/event/openstack/requests/list_events.rb
@@ -18,7 +18,7 @@ module Fog
           end
 
           request(
-            :body    => Fog::JSON.encode(data),
+            :query   => data,
             :expects => 200,
             :method  => 'GET',
             :path    => 'events'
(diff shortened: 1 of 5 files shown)


seanhandley:
Thanks @aufi - can you link to the relevant sections of the OpenStack API? I was unaware of this issue.

Also, you mention trouble with "smart firewalls" ? Would love more info there - I'm wondering if they're too smart for their own good!

aufi:
@seanhandley Hi, thanks for comment!

We're still investigating, I will put there more infro when find something relevant.

The "smart firewall" (yeah, maybe not the best word selection) was meant a web application firewall like Apache mod_security, e.g. with https://www.owasp.org/index.php/Category:OWASP_ModSecurity_Core_Rule_Set_Project

aufi:
Closing this PR, seem to not be the real issue and depends on OpenStack telemetry API.

PR #317, opened 2017-10-11
313 heat files parameter

ioggstream:
## What does it do?

Resolve files in yaml template. 

get_file specs are here https://docs.openstack.org/heat/pike/template_guide/hot_spec.html#get-file

### DIFF
--- a/lib/fog/orchestration/openstack/requests/create_stack.rb
+++ b/lib/fog/orchestration/openstack/requests/create_stack.rb
@@ -1,5 +1,3 @@
-require "fog/orchestration/util/recursive_hot_file_loader"
-
 module Fog
   module Orchestration
     class OpenStack
@@ -34,15 +32,14 @@ module Fog
           # Templates should always:
           #  - be strings
           #  - contain URI references instead of relative paths.
-          # Passing :template_url may not work well with `get_files` and remote `type`:
+          # Passing :template_url may not work well with `get_file` and remote `type`:
           #  the python client implementation in shade retrieves from :template_uri
           #  and replaces it with :template.
           #  see https://github.com/openstack-infra/shade/blob/master/shade/openstackcloud.py#L1201
           #  see https://developer.openstack.org/api-ref/orchestration/v1/index.html#create-stack
-          file_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url], options[:files])
-          files = file_resolver.get_files
+          file_resolver = Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url], options[:files])
           options[:template] = file_resolver.template
-          options[:files] = files if files
+          options[:files] = file_resolver.files unless file_resolver.files.empty?
 
           request(
             :expects => 201,
@@ -90,9 +87,8 @@ module Fog
           end
 
           if options.key?(:template) || options.key?(:template_url)
-            file_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url], options[:files])
-            files = file_resolver.get_files
-            response.body['files'] = files if files
+            file_resolver = Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url], options[:files])
+            response.body['files'] = file_resolver.files unless file_resolver.files.empty?
           end
 
           response
--- a/lib/fog/orchestration/openstack/requests/create_stack.rb
+++ b/lib/fog/orchestration/openstack/requests/create_stack.rb
@@ -1,4 +1,4 @@
-require "fog/orchestration/util/recursive_hot_file_loader"  # FIXME is there a better way to require this file?
+require "fog/orchestration/util/recursive_hot_file_loader"
 
 module Fog
   module Orchestration
@@ -12,6 +12,7 @@ module Fog
         #   * :template [String] Structure containing the template body.
         #   or (one of the two Template parameters is required)
         #   * :template_url [String] URL of file containing the template body.
+        #   * :files [Hash] Hash with files resources.
         #   * :disable_rollback [Boolean] Controls rollback on stack creation failure, defaults to false.
         #   * :parameters [Hash] Hash of providers to supply to template
         #   * :timeout_mins [Integer] Minutes to wait before status is set to CREATE_FAILED
@@ -38,9 +39,9 @@ module Fog
           #  and replaces it with :template.
           #  see https://github.com/openstack-infra/shade/blob/master/shade/openstackcloud.py#L1201
           #  see https://developer.openstack.org/api-ref/orchestration/v1/index.html#create-stack
-          hot_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url])
-          files = hot_resolver.get_files()
-          options[:template] = hot_resolver.template
+          file_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url], options[:files])
+          files = file_resolver.get_files
+          options[:template] = file_resolver.template
           options[:files] = files if files
 
           request(
@@ -85,12 +86,12 @@ module Fog
           }
 
           if options.key?(:files)
-            response.body['files'] = { 'foo.sh' => 'hello' }
+            response.body['files'] = {'foo.sh' => 'hello'}
           end
 
-          if options.key?(:template) or options.key?(:template_url)
-            hot_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url])
-            files = hot_resolver.get_files()
+          if options.key?(:template) || options.key?(:template_url)
+            file_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url], options[:files])
+            files = file_resolver.get_files
             response.body['files'] = files if files
           end
 
--- a/lib/fog/orchestration/openstack/requests/create_stack.rb
+++ b/lib/fog/orchestration/openstack/requests/create_stack.rb
@@ -1,3 +1,5 @@
+require "fog/orchestration/util/recursive_hot_file_loader"  # FIXME is there a better way to require this file?
+
 module Fog
   module Orchestration
     class OpenStack
@@ -27,6 +29,19 @@ module Fog
               :stack_name => arg1
             }.merge(arg2.nil? ? {} : arg2)
           end
+
+          # Templates should always:
+          #  - be strings
+          #  - contain URI references instead of relative paths.
+          # Passing :template_url may not work well with `get_files` and remote `type`:
+          #  the python client implementation in shade retrieves from :template_uri
+          #  and replaces it with :template.
+          #  see https://github.com/openstack-infra/shade/blob/master/shade/openstackcloud.py#L1201
+          #  see https://developer.openstack.org/api-ref/orchestration/v1/index.html#create-stack
+          hot_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url])
+          files = hot_resolver.get_files()
+          options[:template] = hot_resolver.template
+          options[:files] = files if files
 
           request(
             :expects => 201,
@@ -68,6 +83,17 @@ module Fog
             'id'    => stack_id,
             'links' => [{"href" => "http://localhost:8004/v1/fake_tenant_id/stacks/#{options[:stack_name]}/#{stack_id}", "rel" => "self"}]
           }
+
+          if options.key?(:files)
+            response.body['files'] = { 'foo.sh' => 'hello' }
+          end
+
+          if options.key?(:template) or options.key?(:template_url)
+            hot_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url])
+            files = hot_resolver.get_files()
+            response.body['files'] = files if files
+          end
+
           response
         end
       end
--- a/lib/fog/orchestration/openstack/requests/update_stack.rb
+++ b/lib/fog/orchestration/openstack/requests/update_stack.rb
@@ -1,3 +1,5 @@
+require "fog/orchestration/util/recursive_hot_file_loader"
+
 module Fog
   module Orchestration
     class OpenStack
@@ -10,6 +12,7 @@ module Fog
         #   or (one of the two Template parameters is required)
         #   * :template_url [String] URL of file containing the template body.
         #   * :parameters [Hash] Hash of providers to supply to template.
+        #   * :files [Hash] Hash with files resources.
         #
         def update_stack(arg1, arg2 = nil, arg3 = nil)
           if arg1.kind_of?(Stack)
@@ -36,9 +39,9 @@ module Fog
           #  and replaces it with :template.
           #  see https://github.com/openstack-infra/shade/blob/master/shade/openstackcloud.py#L1201
           #  see https://developer.openstack.org/api-ref/orchestration/v1/index.html#create-stack
-          hot_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url])
-          files = hot_resolver.get_files()
-          options[:template] = hot_resolver.template
+          file_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url], options[:files])
+          files = file_resolver.get_files
+          options[:template] = file_resolver.template
           options[:files] = files if files
 
           request(
@@ -66,6 +69,16 @@ module Fog
             options = {
               :stack_name => stack_name
             }.merge(arg3.nil? ? {} : arg3)
+          end
+
+          if options.key?(:files)
+            response.body['files'] = {'foo.sh' => 'hello'}
+          end
+
+          if options.key?(:template) || options.key?(:template_url)
+            file_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url], options[:files])
+            files = file_resolver.get_files
+            response.body['files'] = files if files
           end
 
           response = Excon::Response.new
--- a/lib/fog/orchestration/openstack/requests/update_stack.rb
+++ b/lib/fog/orchestration/openstack/requests/update_stack.rb
@@ -28,6 +28,19 @@ module Fog
             }.merge(arg3.nil? ? {} : arg3)
           end
 
+          # Templates should always:
+          #  - be strings
+          #  - contain URI references instead of relative paths.
+          # Passing :template_url may not work well with `get_files` and remote `type`:
+          #  the python client implementation in shade retrieves from :template_uri
+          #  and replaces it with :template.
+          #  see https://github.com/openstack-infra/shade/blob/master/shade/openstackcloud.py#L1201
+          #  see https://developer.openstack.org/api-ref/orchestration/v1/index.html#create-stack
+          hot_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url])
+          files = hot_resolver.get_files()
+          options[:template] = hot_resolver.template
+          options[:files] = files if files
+
           request(
             :expects => 202,
             :path    => "stacks/#{stack_name}/#{stack_id}",
--- a/lib/fog/orchestration/openstack/requests/update_stack.rb
+++ b/lib/fog/orchestration/openstack/requests/update_stack.rb
@@ -1,5 +1,3 @@
-require "fog/orchestration/util/recursive_hot_file_loader"
-
 module Fog
   module Orchestration
     class OpenStack
@@ -34,15 +32,14 @@ module Fog
           # Templates should always:
           #  - be strings
           #  - contain URI references instead of relative paths.
-          # Passing :template_url may not work well with `get_files` and remote `type`:
+          # Passing :template_url may not work well with `get_file` and remote `type`:
           #  the python client implementation in shade retrieves from :template_uri
           #  and replaces it with :template.
           #  see https://github.com/openstack-infra/shade/blob/master/shade/openstackcloud.py#L1201
           #  see https://developer.openstack.org/api-ref/orchestration/v1/index.html#create-stack
-          file_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url], options[:files])
-          files = file_resolver.get_files
+          file_resolver = Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url], options[:files])
           options[:template] = file_resolver.template
-          options[:files] = files if files
+          options[:files] = file_resolver.files unless file_resolver.files.empty?
 
           request(
             :expects => 202,
@@ -76,9 +73,8 @@ module Fog
           end
 
           if options.key?(:template) || options.key?(:template_url)
-            file_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url], options[:files])
-            files = file_resolver.get_files
-            response.body['files'] = files if files
+            file_resolver = Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url], options[:files])
+            response.body['files'] = file_resolver.files unless file_resolver.files.empty?
           end
 
           response = Excon::Response.new
--- a/lib/fog/orchestration/util/recursive_hot_file_loader.rb
+++ b/lib/fog/orchestration/util/recursive_hot_file_loader.rb
@@ -23,22 +23,14 @@ module Fog
         attr_reader :template
 
         def initialize(template, files = nil)
-          # Serialize the template hash to deep-copy it and
-          #  avoid modifying the argument. Instead create a
-          #  new one to be modified by get_file_contents.
-          #
           # According to https://github.com/fog/fog-openstack/blame/master/docs/orchestration.md#L122
           #  templates can be either String or Hash.
-          @template = deep_copy(template)
-          @files = files || {}
+          #  If it's an Hash, we deep_copy it so the passed argument
+          #  is not modified by get_file_contents.
+          template = deep_copy(template)
           @visited = Set.new
-        end
-
-        def get_files
-          Fog::Logger.debug("Processing template #{@template}")
-          @template = get_template_contents(@template)
-          Fog::Logger.debug("Template processed. Populated #{@files}")
-          @files
+          @files = files || {}
+          @template = get_template_contents(template)
         end
 
         # Return string
@@ -64,9 +56,9 @@ module Fog
         # XXX: we could use named parameters
         # and better mimic heatclient implementation.
         def get_template_contents(template_file)
-          Fog::Logger.debug("get_template_contents #{template_file}")
+          Fog::Logger.debug("get_template_contents [#{template_file}]")
 
-          raise "template_file should be Hash or String" unless
+          raise "template_file should be Hash or String, not #{template_file.class.name}" unless
             template_file.kind_of?(String) || template_file.kind_of?(Hash)
 
           local_base_url = url_join("file:/", File.absolute_path(Dir.pwd))
@@ -100,16 +92,13 @@ module Fog
         #   replaced with their absolute URI as done in heatclient
         #   and shade.
         #
-        def get_file_contents(from_data, base_url = nil)
+        def get_file_contents(from_data, base_url)
           Fog::Logger.debug("Processing #{from_data} with base_url #{base_url}")
 
-          # Recursively traverse the tree.
-          if recurse_if(from_data)
-            recurse_data = from_data.kind_of?(Hash) ? from_data.values : from_data
-            recurse_data.each do |value|
-              get_file_contents(value, base_url)
-            end
-          end
+          # Recursively traverse the tree
+          #   if recurse_data is Array or Hash
+          recurse_data = from_data.kind_of?(Hash) ? from_data.values : from_data
+          recurse_data.each { |value| get_file_contents(value, base_url) } if recurse_data.kind_of?(Array)
 
           # I'm on a Hash, process it.
           return unless from_data.kind_of?(Hash)
@@ -155,17 +144,18 @@ module Fog
             # Validate URI to protect from open-uri attacks.
             url = URI(uri_or_filename)
 
-            # Remote schemes must contain an host.
-            raise ArgumentError if url.host.nil? && remote_schemes.include?(url.scheme)
+            if remote_schemes.include?(url.scheme)
+              # Remote schemes must contain an host.
+              raise ArgumentError if url.host.nil?
 
-            # Encode URI with spaces.
-            uri_or_filename = URI.encode(URI.decode(URI(uri_or_filename).to_s))
+              # Encode URI with spaces.
+              uri_or_filename = uri_or_filename.gsub(/ /, "%20")
+            end
           rescue URI::InvalidURIError
             raise ArgumentError, "Not a valid URI: #{uri_or_filename}"
           end
 
           # TODO: A future revision may implement a retry.
-          # TODO: A future revision may limit download size.
           content = ''
           # open-uri doesn't open "file:///" uris.
           uri_or_filename = uri_or_filename.sub(/^file:/, "")
@@ -202,11 +192,6 @@ module Fog
                          !value.end_with?('.yaml', '.template')
 
           false
-        end
-
-        # Return true if I should inspect this yaml template branch.
-        def recurse_if(value)
-          value.kind_of?(Hash) || value.kind_of?(Array)
         end
 
         # Returns the string baseurl of the given url.
--- a/lib/fog/orchestration/util/recursive_hot_file_loader.rb
+++ b/lib/fog/orchestration/util/recursive_hot_file_loader.rb
@@ -12,29 +12,22 @@ module Fog
       #  a files Hash conforming to Heat Specs
       #  https://developer.openstack.org/api-ref/orchestration/v1/index.html?expanded=create-stack-detail#stacks
       #
-      # Files present in :files are not processed further. The others
-      #   are added to the Hash. This behavior is the same implemented in openstack-infra/shade
-      #   see https://github.com/openstack-infra/shade/blob/1d16f64fbf376a956cafed1b3edd8e51ccc16f2c/shade/openstackcloud.py#L1200
-
-      #
       # This implementation just process nested templates but not resource
       #  registries.
       class RecursiveHotFileLoader
         attr_reader :files
         attr_reader :template
-        attr_reader :template_ori
         attr_reader :visited
-        attr_reader :max_files_size
 
         def initialize(template, files = nil)
           @template = template
-          @template_ori = template
-          @template_base_url = nil
           @files = files || {}
           @visited = Set.new
         end
 
         def get_files
+          return @files unless @files.empty?
+
           Fog::Logger.debug("Processing template #{@template}")
           @template = get_template_contents(@template)
           Fog::Logger.debug("Template processed. Populated #{@files}")
@@ -44,11 +37,12 @@ module Fog
         # Return string
         def url_join(prefix, suffix)
           if prefix
-            prefix += '/' if prefix && !prefix.to_s.end_with?("/")
+            # URI.join replaces prefix parts before a
+            #  trailing slash. See https://docs.ruby-lang.org/en/2.3.0/URI.html.
+            prefix += '/' unless prefix.to_s.end_with?("/")
             suffix = URI.join(prefix, suffix)
             # Force URI to use traditional file scheme representation.
             suffix.host = "" if suffix.scheme == "file"
-            suffix = suffix.to_s
           end
           suffix.to_s
         end
@@ -68,7 +62,7 @@ module Fog
           raise "template_file should be Hash or String" unless
             template_file.kind_of?(String) || template_file.kind_of?(Hash)
 
-          local_base_url = base_url_for_url(normalise_file_path_to_url(Dir.pwd + "/TEMPLATE"))
+          local_base_url = url_join("file:/", File.absolute_path(Dir.pwd))
 
           if template_file.kind_of?(Hash)
             template_base_url = local_base_url
@@ -82,7 +76,7 @@ module Fog
           elsif template_is_url?(template_file)
             template_file = normalise_file_path_to_url(template_file)
             template_base_url = base_url_for_url(template_file)
-            raw_template = get_content(template_file)
+            raw_template = read_uri(template_file)
 
             Fog::Logger.debug("Template visited: #{@visited}")
             @visited.add(template_file)
@@ -123,10 +117,8 @@ module Fog
             str_url = url_join(base_url, value)
 
             next if @files.key?(str_url)
-            # Don't process file:// outside our base_url. TODO raise an exception here?
-            next if file_outside_base_url?(base_url, str_url)
 
-            file_content = get_content(str_url)
+            file_content = read_uri(str_url)
 
             # get_file should not recurse hot templates.
             if key == "type" && template_is_raw?(file_content) && !@visited.include?(str_url)
@@ -153,7 +145,7 @@ module Fog
         # Protect open-uri from malign arguments like
         #  - "|ls"
         #  - multiline strings
-        def get_content(uri_or_filename)
+        def read_uri(uri_or_filename)
           remote_schemes = %w[http https ftp]
           Fog::Logger.debug("Opening #{uri_or_filename}")
 
@@ -193,14 +185,6 @@ module Fog
           true
         rescue ArgumentError, URI::InvalidURIError
           false
-        end
-
-        def file_outside_base_url?(base_url, str_url)
-          ret = base_url && str_url.start_with?("file://") && !str_url.start_with?(base_url)
-          if ret
-            Fog::Logger.debug("Trying to reference a file outside #{base_url}: #{str_url}")
-          end
-          ret
         end
 
         # Return true if I should I process this this file.
--- a/lib/fog/orchestration/util/recursive_hot_file_loader.rb
+++ b/lib/fog/orchestration/util/recursive_hot_file_loader.rb
@@ -1,8 +1,8 @@
+require 'set'
 require 'yaml'
 require 'open-uri'
 require 'objspace'
 require 'fog/core'
-require 'set'
 
 module Fog
   module Orchestration
@@ -11,6 +11,11 @@ module Fog
       # Resolve get_file resources found in a HOT template populating
       #  a files Hash conforming to Heat Specs
       #  https://developer.openstack.org/api-ref/orchestration/v1/index.html?expanded=create-stack-detail#stacks
+      #
+      # Files present in :files are not processed further. The others
+      #   are added to the Hash. This behavior is the same implemented in openstack-infra/shade
+      #   see https://github.com/openstack-infra/shade/blob/1d16f64fbf376a956cafed1b3edd8e51ccc16f2c/shade/openstackcloud.py#L1200
+
       #
       # This implementation just process nested templates but not resource
       #  registries.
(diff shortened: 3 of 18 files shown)


lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,221 @@
...
+          #   with a suitable error.
+          #
+          # XXX Implement a retry here?
+          content = ''
+          # does ruby close the socket/filedescriptor at exit?
+          # XXX Limit download file size
+          # XXX Protect from vanilla open-uri attacks
+          uri_or_filename = uri_or_filename[7..-1] if uri_or_filename.start_with?("file:///")
+          open(uri_or_filename) { |f| content = f.read }
+          content == "Error" ? nil : content
```
bzwei: This line is problematic. You are not assigning the result to anything. Besides, can `content` ever be `Error`?
You made a few comments but did not provide solution. It does not look like final code to be merged.

lib/fog/orchestration/openstack/requests/create_stack.rb
```diff
@@ -28,6 +30,14 @@ def create_stack(arg1, arg2 = nil)
             }.merge(arg2.nil? ? {} : arg2)
           end
 
+          # Eventually resolve files and update original template.
+          hot_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template])
```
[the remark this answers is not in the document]
ioggstream: Yes, I'll add support for template_url too. I need to add this to https://github.com/fog/fog-openstack/blob/master/docs/orchestration.md as it's not mentioned there.

lib/fog/orchestration/openstack/requests/create_stack.rb
```diff
@@ -28,6 +30,14 @@ def create_stack(arg1, arg2 = nil)
...
+          files = hot_resolver.get_files()
+          if files
+            options['files'] = files
+            options['template'] = hot_resolver.template
```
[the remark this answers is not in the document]
ioggstream: To ensure unicity of template resources, we need to:

  - convert every `type` and `get_file` reference to a full URI (eg. file:/// or http:///)
  
On Newton I was able to create the stack referenced in the MIQ ticket using that code, but I still need to write a proper functional testsuite.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,221 @@
...
+
+          @template = template
+          @template_base_url = nil
+          @files = {}
+          @max_files_size = (128 * 1 << 10)
+          @visited = {}
+        end
+
+        def get_files
+          Fog::Logger.warning("Processing template #{@template}")
```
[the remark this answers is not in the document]
ioggstream: Until in [WIP] I need warnings to show those logs. Then I'll move to debug.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: I've been told to create a  [WIP] pull request to simplify code reviews.

For these lines I'd appreciate some help: 

  - which are some common test cases involving open-uri?
  - do you have any suggestions about securing other parts of this code?

ioggstream:
Hi @bzwei Thank you very much for your time!

Consider this patch is [WIP] and has been pushed to collect code-reviews and comments.

Thanks for all the ruby-related comments! If you have some implementation-related feedback or questions please let me know!

bzwei:
@ioggstream Sorry, I did not realize that it is still wip. Take your time to refine and complete. I will review it more.

ioggstream:
@bzwei don't worry, and thank you very much for all your time! You're welcome to suggest whatever you like so that I can fine-tune the code as soon as possible!

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,254 @@
...
+                  raise template_file_exc
+              end
+          end
+        end
+
+        def get_files
+          Fog::Logger.warning("Processing template #{@template}")
+          _, @template = get_template_contents(@template)
+          Fog::Logger.warning("Template processed. Populated #{@files}")
+          return @files
```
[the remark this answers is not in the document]
ioggstream: @seanhandley do I always need to use implicit return in ruby?

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,254 @@
...
+              template_base_url = base_url_for_url(normalise_file_path_to_url(Dir.pwd+"/TEMPLATE"))
+            else
+              template_file = normalise_file_path_to_url(template_file)
+              template_base_url = base_url_for_url(template_file)
+              # TODO: normalize template_file
+              tpl = get_content(template_file)
+              @visited[template_file] = true
+              Fog::Logger.warning("Template visited: #{@visited}")
+            end
+            template = YAML.load(tpl)
```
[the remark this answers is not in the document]
ioggstream: @seanhandley As of now, `YAML.safe_load` has issues with dates. Elsewhere, fog still uses `YAML.load`

```
YAML.safe_load("heat_template_version: 2016-10-14")  # Psych::DisallowedClass exception (Date)
```

See https://github.com/ruby/psych/issues/262#issuecomment-166964077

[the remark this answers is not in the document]
bzwei: Try `YAML.safe_load(tpl, [Date])`

bzwei:
@ioggstream What if downloading a resource file fails, or a local file does not exist? Should we fail the stack creation immediately because anyway it will fail at the provider side, right?

lib/fog/orchestration/util/recursive_hot_file_loader.rb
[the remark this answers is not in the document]
ioggstream: @bzwei ok with `YAML.safe_load(tpl, [Date])`. It shouldn't break things, right? (eg. deserializes and serializes date in the same format, ...)

ioggstream:
@bzwei If a resource file can't be loaded, fog should log and raise an exception. Otherwise, heat will start processing the stack and throw the error in the response.

I'm trying to test the patch inside manageiq and I'm now on both freenode#manageiq and gitter#manageiq. Ping me if you like.

ioggstream:
Update gem in manageiq
```
  git clone https://github.com/ioggstream/fog-openstack.git
  git checkout remotes/origin/313-heat-files-parameter
  cd /var/www/miq/vmdb
  export PATH+=:$PWD/bin
  bundle exec rake install
  bundle exec rake evm:start
```

lib/fog/orchestration/openstack/requests/create_stack.rb
```diff
@@ -28,6 +30,19 @@ def create_stack(arg1, arg2 = nil)
...
 
+          # Templates should always:
+          #  - be strings
+          #  - contain URI references instead of relative paths.
+          # Passing :template_url may not work well with `get_files` and remote `type`:
+          #  the python client implementation in shade retrieves from :template_uri
+          #  and replaces it with :template.
+          #  see https://github.com/openstack-infra/shade/blob/master/shade/openstackcloud.py#L1201
+          #  see https://developer.openstack.org/api-ref/orchestration/v1/index.html#create-stack
+          hot_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url])
```
bzwei: suggest `file_resolver` to better describe the purpose of the variable

bzwei: Use the resolver only when `options[:files]` does not exist. In other words, no need to load files if the caller already provides them. Update the comments in https://github.com/fog/fog-openstack/pull/317/files#diff-360fd267b1a87d726bdd369c23976486R10

lib/fog/orchestration/openstack/requests/update_stack.rb
```diff
@@ -17,22 +17,35 @@ def update_stack(arg1, arg2 = nil, arg3 = nil)
...
 
+          # Templates should always:
+          #  - be strings
+          #  - contain URI references instead of relative paths.
+          # Passing :template_url may not work well with `get_files` and remote `type`:
+          #  the python client implementation in shade retrieves from :template_uri
+          #  and replaces it with :template.
+          #  see https://github.com/openstack-infra/shade/blob/master/shade/openstackcloud.py#L1201
+          #  see https://developer.openstack.org/api-ref/orchestration/v1/index.html#create-stack
+          hot_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url])
```
bzwei: Same comment as for `create_stack`

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,227 @@
...
+  module Orchestration
+    module Util
+      #
+      # Resolve get_file resources found in a HOT template populating
+      #  a files Hash conforming to Heat Specs
+      #  https://developer.openstack.org/api-ref/orchestration/v1/index.html?expanded=create-stack-detail#stacks
+      #
+      # This implementation just process nested templates but not resource
+      #  registries.
+      class RecursiveHotFileLoader
```
bzwei: The logic how to traverse nodes and load files seems over complicated. I have hard time to follow. Can you add comments to each method? We can remove some of them if they turn out to be straightforward.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,254 @@
...
+          # XXX Limit download file size
+          # XXX Protect from vanilla open-uri attacks
+          uri_or_filename = uri_or_filename[7..-1] if uri_or_filename.start_with?("file:///")
+          open(uri_or_filename) { |f| content = f.read }
+          content == "Error" ? nil : content
+
+          content
+        end
+
+        def get_template_contents(template_file, _template_url = nil, _template_object = nil, _existing = false)
```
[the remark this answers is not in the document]
bzwei: Remove the last three arguments if they are not used at all.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,254 @@
...
+          elsif template_file.kind_of?(Hash)
+            # Normalize yaml.
+            Fog::Logger.warning("Reingest")
+            template = YAML.safe_load(YAML.dump(template_file))
+            template_base_url = base_url_for_url(normalise_file_path_to_url(Dir.pwd+"/TEMPLATE"))
+          else
+            raise NotImplementedError, "template should be Hash or String"
+          end
+
+          get_file_contents(template, base_url = template_base_url)
```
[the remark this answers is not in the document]
bzwei: The way to call the method looks strange. Like the bot says, remove `base_url`

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,227 @@
...
+              # TODO: normalize template_file
+              tpl = get_content(template_file)
+              @visited[template_file] = true
+              Fog::Logger.warning("Template visited: #{@visited}")
+            end
+            template = YAML.load(tpl)
+          elsif template_file.kind_of?(Hash)
+            # Normalize yaml.
+            Fog::Logger.warning("Reingest")
+            template = YAML.safe_load(YAML.dump(template_file))
```
bzwei: Why do you need to normalize yaml? Here `template_file` is already a hash, isn't it?

lib/fog/orchestration/openstack/requests/create_stack.rb
ioggstream: @bzwei in shade they merge passed and retrieved :files. Without the retrieved ones the stack will fail.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: @bzwei agree but:

- my first implementation was simpler
- then I preferred to mimic the more stable python implementation

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,232 @@
+require 'yaml'
+require 'open-uri'
+require 'objspace'
+require 'fog/core'
+require 'set'
+
+# Ruby URI is not round-trip safe when schema == file
+#  eg. URI("file:///a.out").to_s != file:///a.out"
+module URI
```
ioggstream: @bzwei  I need to extend the URI module to fix round-trip safety. Is that fine? Should I put that class elsewhere?

ioggstream:
@bzwei  Here is the functional test. There's a commented script on how to run this.
 https://gist.github.com/ioggstream/545e39c13e3a922405326af4f2b211cb

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,236 @@
...
+          # Retrive the content of a local or remote file.
+          #
+          # @param A local or remote uri.
+          #
+          Fog::Logger.debug("Opening #{uri_or_filename}")
+          uri_or_filename = uri_or_filename.to_s if uri_or_filename.kind_of?(URI)
+          # throw exceptions enables stack creation to fail
+          #   with a suitable error.
+          #
+          # XXX Implement a retry here?
```
ioggstream: This first implementation won't contain a retry. It could be added in the future.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,236 @@
...
+          content = ''
+          # does ruby close the socket/filedescriptor at exit?
```
ioggstream: The used syntax 
```
open() { |f| ... }
```
 is expected to play nicely with sockets/filedescriptor.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,236 @@
...
+            template_base_url = local_base_url
+          elsif template_is_url?(template_file)
+            template_file = normalise_file_path_to_url(template_file)
+            template_base_url = base_url_for_url(template_file)
+            raw_template = get_content(template_file)
+
+            Fog::Logger.warning("Template visited: #{@visited}")
+            @visited[template_file] = true
+          else
+            raise NotImplementedError, "template_file is not a string of the expected form"
```
ioggstream: Can I raise ArgumentError here too?

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,236 @@
...
+        def initialize(template, files = nil)
+          @template = template
+          @template_ori = template
+          @template_base_url = nil
+          @files = files || {}
+          @max_files_size = (128 * 1 << 10)
+          @visited = {}
+        end
+
+        def yaml_load(content)
```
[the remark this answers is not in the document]
ioggstream: afaik YAML is a superset of JSON.

```
YAML.load('{"a":"b"}') == YAML.load("a: b")
```

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,236 @@
...
+          return YAML.load(content) if RUBY_VERSION < "2.1"
+
+          YAML.safe_load(content, [Date])
+        end
+
+        def get_files
```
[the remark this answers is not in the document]
ioggstream: @bzwei How can I test private methods?

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,236 @@
...
+          Fog::Logger.warning("Processing template #{@template}")
+          _, @template = get_template_contents(@template)
```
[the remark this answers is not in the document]
ioggstream: if @template is a String, `get_template_contents` can't modify it. We need to always replace it there.

Removing the assignment, the proper unittest fails.

```
Fog::Orchestration[:openstack] | stack requests::success#test_0005_#dont_modify_passed_template [/code/test/requests/orchestration/stack_files_util_tests.rb:89]:

      # The template argument should be modified.
      assert(template['resources']['a_file']['type'].start_with?('file:///'), file_resolver.template)
      ^^^ this will fail.

      # No side effect on the original template.
      refute(@local_yaml['resources']['a_file']['type'].start_with?('file:///'))

```

lib/fog/orchestration/util/recursive_hot_file_loader.rb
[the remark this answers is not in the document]
bzwei: with rspec you test private method as if it is public

lib/fog/orchestration/util/recursive_hot_file_loader.rb
bzwei: Neither is close to what you mean. We can simply raise a string error
```
  raise "template_file is not ..."
```

lib/fog/orchestration/util/recursive_hot_file_loader.rb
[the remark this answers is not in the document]
ioggstream: @bzwei fog-openstack uses minitest, which  seems not doing the job. 

Is it worth to replace all calls with the `send` counterpart?

```
        file_resolver.send(:get_file_contents, data)
```

Otherwise I'll:

  - I'll leave as public the principal template resolution methods (eg. may be useful in further development, like adopting stacks, resource_registry, ..)
  - move private methods to a different file.rb
  - we may add or ignore tests for private methods (imho they will be useful for troubleshooting/debugging)
  
Let me know and I'll do ;)

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,250 @@
...
+        #          - a raw_template string
+        #          - an URI string
+        #          - an Hash containing the parsed template.
+        #
+        # XXX: after deprecation of Ruby 1.9 we could use
+        #      named parameters and better mimic heatclient implementation.
+        def get_template_contents(template_file)
+          Fog::Logger.warning("get_template_contents #{template_file}")
+
+          raise "template_file should be Hash or String" unless
```
bzwei: This is private method for internal use. We may not need to validate the argument all the time.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,250 @@
...
+            Fog::Logger.warning("Template visited: #{@visited}")
+            @visited[template_file] = true
+          else
+            raise "template_file is not a string of the expected form"
+          end
+          template = yaml_load(raw_template)
+
+          get_file_contents(template, template_base_url)
+
+          return nil, template
```
bzwei: Why return `nil, template`. You can return only `template`, which is expected from the method name too.

First of all, I don't understand in what scenario can the input be a Hash.
Secondly, if the input is already a Hash, you don't need to convert it to `raw_template` and then convert it back to Hash.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,250 @@
...
+              end
+
+              @files[str_url] = file_content
+              # replace the data value with the normalised absolute URL
+              Fog::Logger.debug("Replacing #{key} with #{str_url} in #{from_data}")
+              from_data[key] = str_url
+            end
+          end
+        end
+        private :file_outside_base_url?
```
bzwei: You can have a keyword `private` and place all method definitions below this line

```
private

def file_outside_base_url?
end

def get_content
end

...

```

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,250 @@
...
+        #   Resource referenced by get_file and type are eventually
+        #   replaced with their absolute URI as done in heatclient
+        #   and shade.
+        #
+        def get_file_contents(from_data, base_url = nil)
+          Fog::Logger.warning("Processing #{from_data} with base_url #{base_url}")
+
+          # Recursively traverse the tree.
+          if recurse_if(from_data)
+            recurse_data = from_data.kind_of?(Hash) ? from_data.to_a : from_data
```
bzwei: should `from_data.ta_a` be `from_data.values`?

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: @bzwei  `get_template_file` receives the template directly from `get_files`.

Validating here make the code safer and allows to reuse this method for `resource_registry` & co.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: @bzwei 

- removed `nil`
- you can either pass a template as an Hash or a String
- I round-trip the Hash to deep-copy it in another object and don't modifying the passed argument. See https://github.com/ioggstream/fog-openstack/blob/218ee9b905b8a2e265c1e3bdba65e55da01db647/test/requests/orchestration/stack_files_util_tests.rb#L106

lib/fog/orchestration/util/recursive_hot_file_loader.rb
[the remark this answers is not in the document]
ioggstream: done.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -169,6 +76,9 @@ def get_template_contents(template_file)
 
           if template_file.kind_of?(Hash)
             template_base_url = local_base_url
+            # Serialize the template hash to deep-copy it and
```
ioggstream: @bzwei Explain why serialize and deserialize Hash template.

ioggstream:
Close and reopen with 1.9 removal.

ioggstream:
Close and reopen for CI.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,243 @@
...
+        #
+        # @param [String] An heat template key
+        #
+        def ignore_if(key, value)
+          return true if key != 'get_file' && key != 'type'
+
+          return true unless value.kind_of?(String)
+
+          return true if key == 'type' &&
+                         !value.end_with?('.yaml', '.template')
```
bzwei: can it end with '.yml'? should it be case insensitive?

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,243 @@
...
+              next if ignore_if(key, value)
+              Fog::Logger.debug("Inspecting #{key}, #{value} at #{base_url}")
+
+              # Resolve relative paths.
+              base_url += '/' if base_url && !base_url.end_with?('/')
+              str_url = url_join(base_url, value)
+
+              next if @files.key?(str_url)
+              # Don't process file:// outside our base_url. TODO raise an exception here?
+              next if file_outside_base_url?(base_url, str_url)
```
bzwei: I don't understand what is a base_url, and what is a outside_base_url. Can you give me an example and why should we skip it?

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,243 @@
...
+          uri_or_filename = uri_or_filename.sub(/^file:/, "")
+
+          open(uri_or_filename) { |f| content = f.read }
+          content
+        end
+
+        # Return true if the file is an heat template, false otherwise.
+        def template_is_raw?(content)
+          htv = content.index("heat_template_version:")
+          htv && htv < 5
```
bzwei: better to strip off leading white spaces and test whether content starts with "heat_template_version"

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,243 @@
...
+              next if @files.key?(str_url)
+              # Don't process file:// outside our base_url. TODO raise an exception here?
+              next if file_outside_base_url?(base_url, str_url)
+
+              file_content = get_content(str_url)
+
+              # get_file should not recurse hot templates.
+              if key == "type" && template_is_raw?(file_content) && !(@visited[str_url])
+                template = get_template_contents(str_url)
+                file_content = YAML.dump(template)
```
bzwei: Why do you need to convert the content to a hash and dump it back to yaml? Don't you already get the content?

Why do you say "get_file should not recurse hot templates"? What about the nested hot template contains `get_file`?

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: I don't want to support templates referencing generic files.
```
...
get_file: "file:///etc/passwd"
...
```

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: ### file_content 

If I run
```
stack create --template http://foo.com/openshift.yaml
```
and openshift.yaml contains
```
type: master.yaml
```

I expect the original template should upload to heat 
```
# modified openshift.yaml
type: http://foo.com/master.yaml
```
I'm adding a relevant test for that case.

### `type` vs `get_file`

If a template is passed as a `type' then it's part of the stack, and it's used by heat create resources.

```
resources:
  
  server:
    type: server.yaml
    ...
```

If a template is referenced via `get_file`, then it should be considered just a plain file to be attached to the templates, but is not interpreted by heat itself.

```
resources:

  config_files:
    type: OS::Heat::SoftwareConfig
    properties:
      group: script
      config:
      - get_file: uset_just_like_a_file.yaml

```

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: @bzwei While it could end with yml:

- python implementation uses this more strict policy to avoid accidentally passing files to heat https://github.com/openstack/python-heatclient/blob/master/heatclient/common/template_utils.py#L112

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,244 @@
...
+          # Recursively traverse the tree.
+          if recurse_if(from_data)
+            recurse_data = from_data.kind_of?(Hash) ? from_data.values : from_data
+            recurse_data.each do |value|
+              get_file_contents(value, base_url)
+            end
+          end
+
+          # I'm on a Hash, process it.
+          if from_data.kind_of?(Hash)
```
bzwei: You can write this way
```
return unless from_data.kind_of?(Hash)
from_data.each do |key, value|
  ...
end
```

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,244 @@
...
+              # get_file should not recurse hot templates.
+              if key == "type" && template_is_raw?(file_content) && !(@visited[str_url])
+                template = get_template_contents(str_url)
+                file_content = YAML.dump(template)
+              end
+
+              @files[str_url] = file_content
+              # replace the data value with the normalised absolute URL
+              Fog::Logger.debug("Replacing #{key} with #{str_url} in #{from_data}")
+              from_data[key] = str_url
```
bzwei: Why is it necessary to replace a url or filename with the absolute URL? avoid name conflict?

ioggstream: This is by specs. https://docs.openstack.org/heat/pike/template_guide/hot_spec.html#get-file 

Added reference in the comment.
```
The get_file argument must be a static path or URL and not rely on intrinsic functions like get_param. the Orchestration client does not process intrinsic functions (they are only processed by the Orchestration engine).

```

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,244 @@
...
+          if from_data.kind_of?(Hash)
+            from_data.each do |key, value|
+              next if ignore_if(key, value)
+              Fog::Logger.debug("Inspecting #{key}, #{value} at #{base_url}")
+
+              # Resolve relative paths.
+              base_url += '/' if base_url && !base_url.end_with?('/')
+              str_url = url_join(base_url, value)
+
+              next if @files.key?(str_url)
```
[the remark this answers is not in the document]
ioggstream: Heat only knows the entries we put in it, which are full URI. Moreore relative uris may not be unique (eg. same basename in different directories).

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,244 @@
...
+        attr_reader :template_ori
+        attr_reader :visited
+        attr_reader :max_files_size
+
+        def initialize(template, files = nil)
+          @template = template
+          @template_ori = template
+          @template_base_url = nil
+          @files = files || {}
+          @max_files_size = (128 * 1 << 10)
```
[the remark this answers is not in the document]
ioggstream: Ok, created https://github.com/fog/fog-openstack/issues/334 for future reference.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,241 @@
...
+          Fog::Logger.debug("Processing template #{@template}")
+          @template = get_template_contents(@template)
+          Fog::Logger.debug("Template processed. Populated #{@files}")
+          @files
+        end
+
+        # Return string
+        def url_join(prefix, suffix)
+          if prefix
+            prefix += '/' if prefix && !prefix.to_s.end_with?("/")
```
bzwei: Don't need this line.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,241 @@
...
+            suffix = URI.join(prefix, suffix)
+            # Force URI to use traditional file scheme representation.
+            suffix.host = "" if suffix.scheme == "file"
+            suffix = suffix.to_s
```
bzwei: don't need this line

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: URI.join replaces prefix parts before a trailing slash. See https://docs.ruby-lang.org/en/2.3.0/URI.html. Added comment to explain why this line is required.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: @bzwei I did it to ensure returned value is a string and not an uri.

ioggstream:
recheck

ioggstream:
recheck

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,243 @@
...
+      #   are added to the Hash. This behavior is the same implemented in openstack-infra/shade
+      #   see https://github.com/openstack-infra/shade/blob/1d16f64fbf376a956cafed1b3edd8e51ccc16f2c/shade/openstackcloud.py#L1200
+
+      #
+      # This implementation just process nested templates but not resource
+      #  registries.
+      class RecursiveHotFileLoader
+        attr_reader :files
+        attr_reader :template
+        attr_reader :template_ori
```
bzwei: remove it and `@template_ori` since they are no longer used.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,243 @@
...
+        attr_reader :visited
+        attr_reader :max_files_size
+
+        def initialize(template, files = nil)
+          @template = template
+          @template_ori = template
+          @template_base_url = nil
+          @files = files || {}
```
bzwei: My original recommendation was that if `files` are already given by caller, don't go through this `RecusiveHotFileLoader` at all. My assumption is that the user already resolved `files`, there is no need to read and process the template.

You decided to accept `files` as an argument and always go through `RecusiveHotFileLoader`, which is time consuming but may be useful one one scenario - if `files` is incomplete, and we still want to help discover the rest. But in your code I don't see you treat `files` as a user passed in argument.

We don't need to be this flexible. I would still recommend what I originally recommended.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,243 @@
...
+      class RecursiveHotFileLoader
+        attr_reader :files
+        attr_reader :template
+        attr_reader :template_ori
+        attr_reader :visited
+        attr_reader :max_files_size
+
+        def initialize(template, files = nil)
+          @template = template
+          @template_ori = template
```
bzwei: delete

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,243 @@
...
+          @template_base_url = nil
```
bzwei: delete

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,243 @@
...
+
+      #
+      # This implementation just process nested templates but not resource
+      #  registries.
+      class RecursiveHotFileLoader
+        attr_reader :files
+        attr_reader :template
+        attr_reader :template_ori
+        attr_reader :visited
+        attr_reader :max_files_size
```
bzwei: delete

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,243 @@
...
+      #   see https://github.com/openstack-infra/shade/blob/1d16f64fbf376a956cafed1b3edd8e51ccc16f2c/shade/openstackcloud.py#L1200
+
+      #
+      # This implementation just process nested templates but not resource
+      #  registries.
+      class RecursiveHotFileLoader
+        attr_reader :files
+        attr_reader :template
+        attr_reader :template_ori
+        attr_reader :visited
```
bzwei: delete

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,243 @@
...
+        # Return string
+        def url_join(prefix, suffix)
+          if prefix
+            # URI.join replaces prefix parts before a
+            #  trailing slash. See https://docs.ruby-lang.org/en/2.3.0/URI.html.
+            prefix += '/' unless prefix.to_s.end_with?("/")
+            suffix = URI.join(prefix, suffix)
+            # Force URI to use traditional file scheme representation.
+            suffix.host = "" if suffix.scheme == "file"
+            suffix = suffix.to_s
```
bzwei: delete

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,243 @@
...
+
+        # Return true if it's an URI, false otherwise.
+        def template_is_url?(path)
+          normalise_file_path_to_url(path)
+          true
+        rescue ArgumentError, URI::InvalidURIError
+          false
+        end
+
+        def file_outside_base_url?(base_url, str_url)
```
bzwei: delete method. not used

lib/fog/orchestration/openstack/requests/create_stack.rb
```diff
@@ -28,6 +31,19 @@ def create_stack(arg1, arg2 = nil)
...
+          #  - be strings
+          #  - contain URI references instead of relative paths.
+          # Passing :template_url may not work well with `get_files` and remote `type`:
+          #  the python client implementation in shade retrieves from :template_uri
+          #  and replaces it with :template.
+          #  see https://github.com/openstack-infra/shade/blob/master/shade/openstackcloud.py#L1201
+          #  see https://developer.openstack.org/api-ref/orchestration/v1/index.html#create-stack
+          file_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url], options[:files])
+          files = file_resolver.get_files
+          options[:template] = file_resolver.template
```
bzwei: According to your logic `file_resolver.template` returns a Hash. I assume `options[:template]` expects a string, i.e., the content of the template.

bzwei:
@ioggstream 
class `RecusiveHotFileResolver` has very similar method names that are very confusing. Consider rename them, for example,
`get_file_contents` -> `traverse_template` or `process_template`
`get_template_contents` -> `parse_template_to_hash`, or simply `parse_template`
`get_content` is fine if we name above methods, or rename `read_file`

Another consideration is the performance. We need to fine tune the code to avoid unnecessary conversion between content and hash.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,243 @@
...
+        #
+        # XXX: after deprecation of Ruby 1.9 we could use
+        #      named parameters and better mimic heatclient implementation.
+        def get_template_contents(template_file)
+          Fog::Logger.debug("get_template_contents #{template_file}")
+
+          raise "template_file should be Hash or String" unless
+            template_file.kind_of?(String) || template_file.kind_of?(Hash)
+
+          local_base_url = base_url_for_url(normalise_file_path_to_url(Dir.pwd + "/TEMPLATE"))
```
bzwei: why use a hard-coded path? what is the use?

ioggstream: Simplified ;)

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: @bzwei on this subject, I implemented the same behavior of heatclient so openstack user can expect a similar behavior between CFME and other python tools (Ansible Tower).

I'm moving that code in a different patch.

lib/fog/orchestration/openstack/requests/create_stack.rb
```diff
@@ -28,6 +31,19 @@ def create_stack(arg1, arg2 = nil)
...
 
+          # Templates should always:
+          #  - be strings
+          #  - contain URI references instead of relative paths.
+          # Passing :template_url may not work well with `get_files` and remote `type`:
+          #  the python client implementation in shade retrieves from :template_uri
+          #  and replaces it with :template.
+          #  see https://github.com/openstack-infra/shade/blob/master/shade/openstackcloud.py#L1201
+          #  see https://developer.openstack.org/api-ref/orchestration/v1/index.html#create-stack
+          file_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url], options[:files])
```
bzwei: better this way
```
if options[:files].nil?
  file_resolver = Fog::Orchestration::Util::RecursiveFileLoader.new(options[:template] || options[:template_url])
  files = file_resolver.get_files
  options[:files] = files unless files.empty?
  options[:template] = file_resolver.template
end

File: lib/fog/orchestration/util/recursive_hot_file_loader.rb [hunk 46]
```diff
@@ -0,0 +1,231 @@
...
+      # Resolve get_file resources found in a HOT template populating
+      #  a files Hash conforming to Heat Specs
+      #  https://developer.openstack.org/api-ref/orchestration/v1/index.html?expanded=create-stack-detail#stacks
+      #
+      # This implementation just process nested templates but not resource
+      #  registries.
+      class RecursiveHotFileLoader
+        attr_reader :files
+        attr_reader :template
+        attr_reader :visited
```
bzwei: delete, no need to explore it as read_only attribute

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,231 @@
...
+
+        def initialize(template, files = nil)
```
bzwei:
```
def initialize(template)
```

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,231 @@
...
+          # Serialize the template hash to deep-copy it and
+          #  avoid modifying the argument. Instead create a
+          #  new one to be modified by get_file_contents.
+          @template = deep_copy(template)
```
bzwei: No need to deep_copy. Why do we expect template can be a Hash? Initially it is either the content string or url.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,231 @@
...
+          @files = files || {}
+          @visited = Set.new
+        end
+
+        def get_files
+          return @files unless @files.empty?
```
bzwei:
```
return @files unless @files.nil? # avoid processing templates for every `get_files` call
@files = {}
```

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,231 @@
...
+        attr_reader :files
+        attr_reader :template
+        attr_reader :visited
+
+        def initialize(template, files = nil)
+          # Serialize the template hash to deep-copy it and
+          #  avoid modifying the argument. Instead create a
+          #  new one to be modified by get_file_contents.
+          @template = deep_copy(template)
+          @files = files || {}
```
bzwei: delete

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,231 @@
...
+          @visited = Set.new
+        end
+
+        def get_files
```
bzwei: `get_files` method is not needed if we write `initialize` this way
```
def initialize(template)
  @files = {}
  @visited = Set.new
  @template = get_template_content(template)
end
```
In the caller you can simply
```
file_resolver = Fog::Orchestration::Util::RecursiveHotFileLoader.new(options[:template] || options[:template_url]
options[:template] = file_resolver.template
options[:files] = file_resolver.files unless file_resolver.files.empty?
```

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: According to https://github.com/fog/fog-openstack/blob/master/docs/orchestration.md#stacks
stack_update accepts Hash too.

```
# from the above link.
heat_template = { "template": { "description": "Updated description" } }
stack.save(heat_template)

```
This allows the code to work in both cases.

ioggstream: Ok, we agreed on irc to leave the deepcopy and support Hash to support orchestration.md examples.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: @bzwei agreed on gitter that we'll reintroduce `files` argument.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: @bzwei reintroduced after gitter discussion.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: Reintroduced after gitter discussion

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: - `visited` removed
- `template` is used to return the processed template with URIs

lib/fog/orchestration/openstack/requests/create_stack.rb
ioggstream: :files reintroduced after gitter discussion

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,227 @@
...
+          template
+        end
+
+        # Traverse the template tree looking for get_file and type
+        #   and populating the @files attribute with their content.
+        #   Resource referenced by get_file and type are eventually
+        #   replaced with their absolute URI as done in heatclient
+        #   and shade.
+        #
+        def get_file_contents(from_data, base_url = nil)
```
bzwei: Can `base_url` ever be `nil`? You always have a base_url when calling from `get_template_contents`. We don't need a default value here.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,227 @@
...
+          Fog::Logger.debug("Processing #{from_data} with base_url #{base_url}")
+
+          # Recursively traverse the tree.
+          if recurse_if(from_data)
```
bzwei: We don't need a `recuse_if` method because you need to test again whether `from_data` is a `Hash`.
Consider
```
recurse_data = from_data.kind_of?(Hash) ? from_data.values : from_data
recurse_data.each { |value| get_file_contents(value, base_url) } if recurse_data.kind_of?(Array)
```

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,227 @@
...
+
+          begin
+            # Validate URI to protect from open-uri attacks.
+            url = URI(uri_or_filename)
+
+            # Remote schemes must contain an host.
+            raise ArgumentError if url.host.nil? && remote_schemes.include?(url.scheme)
+
+            # Encode URI with spaces.
+            uri_or_filename = URI.encode(URI.decode(URI(uri_or_filename).to_s))
```
bzwei: Do we need to address the obsolete warning?
you already have `url`, so the line could be `url_or_filename = URI.encode(URI.decode(url.to_s))`

lib/fog/orchestration/util/recursive_hot_file_loader.rb
```diff
@@ -0,0 +1,227 @@
...
+        def normalise_file_path_to_url(path)
+          # Nothing to do on URIs
+          return path if URI(path).scheme
+
+          path = File.absolute_path(path)
+          url_join('file:/', path)
+        end
+
+        def deep_copy(item)
+          return item if item.kind_of?(String)
```
bzwei: better
```
return item unless item.kind_of?(Hash)
```

ioggstream: while imho:
- that won't deep_copy arrays
- may result counter-intuitive during further development

I'm fine for changing if it's a merge-blocker ;)

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: 1- which obsolete warning?

2- Applied your suggestion and removed the 3rd `URI` call

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: Ok.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: I think it is more flexible/readable having a single function to decide, but your implementation is so clever :D

I changed it and added a small comment to make it more straightforward.

lib/fog/orchestration/util/recursive_hot_file_loader.rb
ioggstream: Replaced with gsub(/ /, "%20")

lib/fog/orchestration/openstack/requests/create_stack.rb
```diff
@@ -1,3 +1,5 @@
+require "fog/orchestration/util/recursive_hot_file_loader"
```
bzwei: You can add the package in https://github.com/fog/fog-openstack/blob/master/lib/fog/openstack.rb then you don't need to `require` here and nor need to use the full namespace while using the resolver class.

ioggstream: Rock! Thank you!!!

bzwei:
@ioggstream Thanks!

ioggstream:
@bzwei Thanks to you!

Ladas:
eh, github acts funky

Issue #320, opened 2017-10-13
Fix DEPRECATION: 


ioggstream:
### I expect

the test use the last interface

### Instead I get

```
.[fog][DEPRECATION] #create_stack(stack_name, options) is deprecated, use #create_stack(options) instead (/code/test/requests/orchestration/stack_tests.rb:47:in `block (3 levels) in <top (required)>')
.[fog][DEPRECATION] #update_stack(stack_id, stack_name, options) is deprecated, use #update_stack(stack, options) instead (/code/test/requests/orchestration/stack_tests.rb:59:in `block (3 levels) in <top (required)>')
..[fog][DEPRECATION] #delete_stack(stack_name, stack_id) is deprecated, use #delete_stack(stack) instead (/code/test/requests/orchestration/stack_tests.rb:67:in `block (3 levels) in <top (required)>')
```

PR #321, opened 2017-10-13
#320 fix deprecation warnings in stack_tests

ioggstream:
Fixed deprecation warnings in stack_tests.

As  @stacks variable defined here
 https://github.com/fog/fog-openstack/compare/master...ioggstream:320-fix-deprecation?expand=1#diff-47b26349d7f38ae7fcd7ef799c7d2f0eL47 

is a boolean, I created a new Stack instance.


gildub:
LGTM, thanks for addressing this!

PR #330, opened 2017-10-20
Exclude test methods from BlockLength check

ioggstream:
### I expect

Test blocks to be longer that 25 lines, and generally longer than actual blocks.

### This PR

Excludes test methods from BlockLength check.


gildub:
LGTM

PR #339, opened 2017-11-08
Allow soft affinity server groups

FlorianNachtigall:
OpenStack Mitaka adds support for soft affinity and anti affinty groups.
Using these with an older installation of OpenStack results in an 400
http error code.
Also using Compute version lesser than 2.15 results in the same 400
code.

### DIFF
--- a/lib/fog/compute/openstack/requests/create_server_group.rb
+++ b/lib/fog/compute/openstack/requests/create_server_group.rb
@@ -11,6 +11,7 @@ module Fog
           }}
           request(
             :body    => Fog::JSON.encode(body),
+            :headers => { "OpenStack-API-Version" => "compute latest" },
             :expects => 200,
             :method  => 'POST',
             :path    => 'os-server-groups'
--- a/lib/fog/compute/openstack/requests/create_server_group.rb
+++ b/lib/fog/compute/openstack/requests/create_server_group.rb
@@ -11,7 +11,7 @@ module Fog
           }}
           request(
             :body    => Fog::JSON.encode(body),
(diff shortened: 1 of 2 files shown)


lib/fog/compute/openstack/requests/create_server_group.rb
```diff
@@ -11,6 +11,7 @@ def create_server_group(name, policy)
           }}
           request(
             :body    => Fog::JSON.encode(body),
+            :headers => {'OpenStack-API-Version' => 'compute latest'},
```
voelzmo: I'm not sure this is how it is supposed to be done.
* we probably should not set this to 'latest', as we don't support 'future microversions' yet, but rather use a concrete number
* we shouldn't set this header here, but rather for the complete compute client
* maybe this should be done similar to how it was introduced for the shared_file_system client with commit https://github.com/fog/fog-openstack/commit/f72d2fce5fc0c89f9a901eb97767b095f4b3522a#diff-18ad1af87ba332277f23d6e3b943f086 which added a `SUPPORTED_MICROVERSION = '2.15'.freeze`?

Any input, @Carthaca?

FlorianNachtigall:
This PR is being followed up by: https://github.com/fog/fog-openstack/pull/340
Instead of introducing microversions only for the create server group request, we requested to introduce microversions to the entire compute service (similar as it has been done for the [shared file system](https://github.com/fog/fog-openstack/commit/f72d2fce5fc0c89f9a901eb97767b095f4b3522a#diff-18ad1af87ba332277f23d6e3b943f086)).

PR #340, opened 2017-11-15
Introduce microversions to compute service

FlorianNachtigall:
OpenStack Mitaka adds support for soft (anti) affinity. For this the compute API mircoversion 2.15 is required. Older installation of OpenStack and older API microversions will result in a 400
http error code.

- We introduced Microversions to the compute service in order to support soft (anti) affinity.
- We added the soft affinity policies for server groups.
- We explicitly request either the v2 or v2.1 API endpoint version of the compute service.

Note: This PR is a follow up of https://github.com/fog/fog-openstack/pull/339

### DIFF
--- a/lib/fog/compute/openstack.rb
+++ b/lib/fog/compute/openstack.rb
@@ -1,7 +1,7 @@
 module Fog
   module Compute
     class OpenStack < Fog::Service
-      SUPPORTED_VERSIONS = /v2\.0|v2\.1/.freeze
+      SUPPORTED_VERSIONS = /v2\.0|v2\.1/
       SUPPORTED_MICROVERSION = '2.15'.freeze
 
       requires :openstack_auth_url
--- a/lib/fog/compute/openstack.rb
+++ b/lib/fog/compute/openstack.rb
@@ -1,8 +1,9 @@
-
-
 module Fog
   module Compute
     class OpenStack < Fog::Service
+      SUPPORTED_VERSIONS = /v2\.0|v2\.1/
+      SUPPORTED_MICROVERSION = '2.15'.freeze
+
       requires :openstack_auth_url
       recognizes :openstack_auth_token, :openstack_management_url,
                  :persistent, :openstack_service_type, :openstack_service_name,
@@ -380,6 +381,10 @@ module Fog
         end
 
         def initialize(options = {})
+          @supported_versions = SUPPORTED_VERSIONS
+          @supported_microversion = SUPPORTED_MICROVERSION
+          @microversion_key = 'X-OpenStack-Nova-API-Version'.freeze
+
           initialize_identity options
 
           @openstack_identity_service_type = options[:openstack_identity_service_type] || 'identity'
@@ -391,9 +396,9 @@ module Fog
 
           authenticate
 
-          unless @path =~ /1\.1|v2/
+          unless @path =~ /\/(v2|v2\.0|v2\.1)\//
             raise Fog::OpenStack::Errors::ServiceUnavailable,
-                  "OpenStack compute binding only supports version 2 (a.k.a. 1.1)"
+                  "OpenStack compute binding only supports version v2 and v2.1"
           end
 
           @persistent = options[:persistent] || false
--- a/lib/fog/compute/openstack.rb
+++ b/lib/fog/compute/openstack.rb
@@ -1,7 +1,7 @@
 module Fog
   module Compute
     class OpenStack < Fog::Service
-      SUPPORTED_VERSIONS = /v2\.0|v2\.1/
+      SUPPORTED_VERSIONS = /v2\.0|v2\.1/.freeze
       SUPPORTED_MICROVERSION = '2.15'.freeze
 
       requires :openstack_auth_url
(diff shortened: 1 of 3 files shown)


BeckerMax:
@Carthaca Can you provide us feedback? Is this ready for merging?

Carthaca:
lgtm

Kiemes:
@Carthaca Can you also merge the PR? Or should we ping someone else?

lib/fog/compute/openstack.rb
```diff
@@ -1,8 +1,9 @@
-
-
 module Fog
   module Compute
     class OpenStack < Fog::Service
+      SUPPORTED_VERSIONS = /v2\.0|v2\.1/
```
gildub: Why not `freeze` here too?

lib/fog/compute/openstack.rb
```diff
@@ -380,6 +381,10 @@ def self.not_found_class
         end
 
         def initialize(options = {})
+          @supported_versions = SUPPORTED_VERSIONS
+          @supported_microversion = SUPPORTED_MICROVERSION
+          @microversion_key = 'X-OpenStack-Nova-API-Version'.freeze
```
gildub: Why a `freeze` here?

lib/fog/compute/openstack.rb
```diff
@@ -1,8 +1,9 @@
-
-
 module Fog
   module Compute
     class OpenStack < Fog::Service
+      SUPPORTED_VERSIONS = /v2\.0|v2\.1/
+      SUPPORTED_MICROVERSION = '2.15'.freeze
```
gildub: Why only `2.15`? Not to trigger more questions about how microversions are handled.

lib/fog/compute/openstack.rb
voelzmo: This is taken 1-by-1 from the [PR which introduced microversions](https://github.com/fog/fog-openstack/pull/242) for the shared-filesystem-service. We can totally change it, if you prefer.

lib/fog/compute/openstack.rb
voelzmo: See below. We can change it either way, whatever you prefer. I don't have strong feelings on freezing.

lib/fog/compute/openstack.rb
voelzmo: We're just introducing 2.15 here, because the feature we're adding is available with this microversion. There are two options, I guess: Updating to the currently most recent microversion for nova, or leaving it at 2.15. I'm tending more towards the latter, as I can't verify the changes e.g. against Pike easily. WDYT?

lib/fog/compute/openstack.rb
gildub: Yeah I remember that one although I realize now that freezing the content of a variable might not make sense. Could you remove the freeze here?

lib/fog/compute/openstack.rb
gildub: It's just to be consistent with other constant, I would tend to freeze the regex as well.

lib/fog/compute/openstack.rb
gildub: Well the issue here is the general handling of microversions (let me save my rant about how microversion is not helping SDKs but serve only Openstack APIs themselves!). Well I don't want to block this PR meanwhile I'm not sure this is the right thing to do, and again that's likely completely inherent to the way microversion works (or how it wants us to work with). So unless someone has a better idea, then we can merge and continue to be doomed with microversion.

lib/fog/compute/openstack.rb
voelzmo: sure, we'll do that

lib/fog/compute/openstack.rb
voelzmo: sure, we'll do that
\

lib/fog/compute/openstack.rb
voelzmo: Sorry, I didn't want to trigger bad feelings about microversion handling on the OpenStack side here :(
I'm entirely with you that it is actually pretty cumbersome to deal with this concept on the SDK or CLI level – seems like the easier way out for the API developers.
So, for the sake of this PR, I'd say we stick to microversion 2.15. I assume that the next person who needs a feature which is supported in >2.15 will have to do a similar thing. Looking at https://docs.openstack.org/nova/latest/reference/api-microversion-history.html I feel like depending on the needed microversion, a lot of non-related changes might to be done, though.

Kiemes:
@gildub we updated the PR accordingly to your comments. The houndci did not like the `freeze` on the `SUPPORTED_VERSIONS` constant.

Can you have a look again?

lib/fog/compute/openstack.rb
gildub: @voelzmo,

It seems like we don't have another solution for now and not block this PR.
Looks like we effectively need bumping the micro-version too as the latest is 2.56 but that's a different PR for sure! 
Thanks for joining the conversation, we likely need to continue this on a dedicated issue thread.

voelzmo: I've opened https://github.com/fog/fog-openstack/issues/345 to track some of the issues, so the discussion can continue. Thanks for merging!

voelzmo:
@gildub thanks for merging. Can we get a new release of fog-openstack containing that PR?

PR #347, opened 2018-01-03
Allow compute URLs ending with service version

beyhan:
Some OpenStack systems provide URLs like
http://host:port/v2.1 for the compute service which didn't match.
This fix changes the version regex to match such URLs

### DIFF
--- a/lib/fog/compute/openstack.rb
+++ b/lib/fog/compute/openstack.rb
@@ -396,7 +396,7 @@ module Fog
 
           authenticate
 
-          unless @path =~ %r{/(v2|v2\.0|v2\.1)/}
+          unless @path =~ %r{/(v2|v2\.0|v2\.1)}
             raise Fog::OpenStack::Errors::ServiceUnavailable,
                   "OpenStack compute binding only supports version v2 and v2.1"
           end


voelzmo:
Seems like the travis builds failed due to a different reason than our code changes. Can we re-trigger the checks ourselves somehow?

timuralp:
@voelzmo there should be a "Restart build" button in the top right on Travis. I re-triggered the tests.

voelzmo:
I think 'restart build' is only there for committers. Mere contributors can't see that button.

Carthaca:
close and re-open should work in that case

timuralp:
@voelzmo unfortunately, you're bumping into https://github.com/fog/fog-openstack/issues/349. I submitted a PR to resolve it (#351). I also discovered that there are two more spec tests that fail when running them locally. @Carthaca do you observe failing "Reloads image state from the server" and "CRUD rbacs" tests as well?

I'll try to figure out what's happening with them :\

timuralp:
Debugging this further, it looks like the two failing spec tests arise after upgrading to fog-core 2.0. I'm going to change the gemspec to point to `fog-core ~> 1.40` and we can tackle the upgrade to 2.0 separately.

voelzmo:
thanks for debugging, @timuralp!

voelzmo:
Seems like rebasing on master helped with the travis checks, thanks @timuralp

Carthaca:
lgtm

PR #358, opened 2018-01-29
Fix get_project request path

aufi:
Get project method failed on bad component(expected absolute path component),
caused by question mark in the path. Updating request to same approach as other
working requests.

### DIFF
--- a/lib/fog/openstack/identity/v3/models/projects.rb
+++ b/lib/fog/openstack/identity/v3/models/projects.rb
@@ -33,9 +33,9 @@ module Fog
             load(service.auth_projects(options).body['projects'])
           end
 
-          def find_by_id(id, options = []) # options can include :subtree_as_ids, :subtree_as_list, :parents_as_ids, :parents_as_list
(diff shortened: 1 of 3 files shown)


mansam:
It would be really great to get this merged asap once @aufi fixes the tests :)

aufi:
@Ladas or others Hi, do you know if there are steps how to re-record VCRs? Thanks

gildub:
@aufi, in your local environment delete ' spec/fixtures/openstack/identity_v3/idv3_project_crud.yml' file then run `bundle exec rake spec` and see the error message from VCR.

Basically the `:new_episodes` option would allow you to record a new cassette. For that you would have to add the option in the line:
```ruby
 VCR.use_cassette('authv3_a', :new_episodes)
```
Or something like that.

gildub:
@aufi, there was no need to change VCR, the issue was some `options` still being initialized as Array.

gildub:
@mansam, do know if there are any back port needs?

mansam:
@gildub Not that I am aware of.

PR #361, opened 2018-02-02
Change path logic when openstack identity prefix is 'v2.0'

alexander-demichev:
There is an issue on RHOS 11 with identity service and keystone v2, when creating a tenant. Problem is that, path built by fog looks like this `/v2.0//v3`, when needed path is `/v2.0/`. I believe this PR can fix it.
I would like to mention this PR https://github.com/fog/fog-openstack/pull/99

P.S. In my case identity prefix comes from MiQ [link](https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/lib/manageiq/providers/openstack/legacy/openstack_handle/handle.rb#L201)

ping @aufi

### DIFF
--- a/lib/fog/identity/openstack.rb
+++ b/lib/fog/identity/openstack.rb
@@ -73,6 +73,8 @@ module Fog
             @path = "/#{options[:openstack_identity_prefix]}/#{@path}"
           end
 
+          @path = "/#{options[:openstack_identity_prefix]}/" if options[:openstack_identity_prefix] == "v2.0"
+
           @persistent = options[:persistent] || false
           @connection = Fog::Core::Connection.new("#{@scheme}://#{@host}:#{@port}", @persistent, @connection_options)
         end


alexander-demichev:
ping @Ladas

Ladas:
@alexander-demichev can you find the exact place setting the '/v3' ? We might need to find a more generic fix. Also this breaks a lot of specs it seems.

Ladas:
@alexander-demichev the problem might be here https://github.com/alexander-demichev/fog-openstack/blob/c307ead493ca1d1b68cfdf1f15b53f505fdc6eef/lib/fog/identity/openstack.rb#L26, where we decide if we should build V2 or V3 service

alexander-demichev:
@Ladas So overcloud allows me to use both v2 and v3 for Identity service. Provider is added to MiQ with keystone v2, created fog object is v2 and path is `/v3`.

alexander-demichev:
@petrblaho @aufi Hi, can you test this?

lib/fog/identity/openstack.rb
```diff
@@ -73,6 +73,8 @@ def initialize(options = {})
             @path = "/#{options[:openstack_identity_prefix]}/#{@path}"
           end
 
+          @path = "/#{options[:openstack_identity_prefix]}/" if options[:openstack_identity_prefix] == "v2.0"
```
aufi: This change doesn't look good to me, when there is "v2.0" prefix, then path is dropped. Could you test some API calls (or ManageIQ refresh) first?

gildub:
@alexander-demichev, first off, thanks for tracking that one. Effectively v2.0 needs to be added to the path when using Identity version 2. We got used to it because that was the default. I confirm this works, meanwhile per @ladas comment to make it more generic, that should be added to V2 class, for instance: 

```ruby
index 6529a93..df74d84 100644
--- a/lib/fog/identity/openstack/v2.rb
+++ b/lib/fog/identity/openstack/v2.rb
@@ -169,6 +169,11 @@ module Fog
         end
 
         class Real < Fog::Identity::OpenStack::Real
+         def initialize(options)
+            options[:openstack_identity_prefix] = 'v2.0' unless options[:openstack_identity_prefix]
+            super
+          end
+
           private
 
           def default_service_type(_)
```

lib/fog/identity/openstack.rb
gildub: Yes, you're right it should actually be:
```ruby
 @path = "/#{options[:openstack_identity_prefix]}/#{@path}" if options[:openstack_identity_prefix] == "v2.0"
```

alexander-demichev:
@gildub Hi, thanks for review. I tried solution from previous comment, it's not working for me.

gildub:
@alexander-demichev, I've tested against OSP12 (Pike). Ok, so you're using code from https://github.com/fog/fog-openstack/pull/361#issuecomment-364102591, could you please provide the request details (path) you're obtaining?

alexander-demichev:
@gildub Hi, it works for me on OSP 12, and doesn't work on OSP 11. After using from your comment path becomes `:path          => "/v3/tenants"`
(I'm using keystone V2)
If you need OSP 11 where this bug is reproducible, I can send you credentials.

gildub:
@alexander-demichev, yes please.

gildub:
@alexander-demichev, the reason this doesn't work on the osp11 you provided is because the identity endpoint is setup with "v3":

```ruby
 {"endpoints"=>
  [{"adminURL"=>"http://10.8.196.59:35357/v3",
    "region"=>"RegionOne",
    "internalURL"=>"http://10.8.196.59:5000/v3",
    "id"=>"2c7f39dde4d343668e8a6a202fcfd616",
    "publicURL"=>"http://10.8.196.59:5000/v3"}],
 "endpoints_links"=>[],
 "type"=>"identity",
 "name"=>"keystone"}

```

In order to use "v2.0" with such catalog is to change the endpoint to be version-less  or provide another service endpoint dedicated to v2.

lib/fog/identity/openstack.rb
mansam: The first version is wrong, but that proposed fix seems to make the line identical to the conditional above-- ` @path = "/#{options[:openstack_identity_prefix]}/#{@path}"`

gildub:
To clarify, there are always 2 aspects to the Identity, the first one is the authentication part using `:openstack_auth_url` and the other part is the identity service which is used by requests (i.e. list_tenants or list_projects). 

And combining the 2 is possible, such as authenticating and obtaining a token against Keystone v3 (the default) and interacting with Keystone v2.0.

The issue here is that fog-openstack is "guessing" from the URL path the version of the identity for both the authentication and the identity service. Although that's done separately.

gildub:
@alexander-demichev, I've tried to tackle this one but the outcome is far from ideal, the best result is when identity endpoint doesn't contain the version because otherwise fog-openstack injects the version in the `@path`. That means OSP11 and alike have to have their configuration version-less. This is how OpenStack Keystone is supposed to be configured and therefore I would push that one back to OpenStack OSP11 support. 

Alternatively one could consider Keystone v2 obsolete since it has been deprecated and therefore no fog-openstack user/client should be using it, attended fog-openstack could detect the version of OpenStack its dealing with. 

What do you think? 

@marek, @Ladas  ^

alexander-demichev:
ping @aufi

gildub:
@alexander-demichev, a work around is to change OSP11 identity endpoint to be like OSP12 without the verision is the url.  Could you please try it? 
I'm still working a solution which implies some structural changes which are quite heavy and therefore take times.

gildub:
I'm closing this issue since it's happening because Keystone endpoint is wrongly set-up to "http://10.8.196.59:35357/v3" where it should be "http://10.8.196.59:35357". 

That said fog-openstack should be able to handle such scenario. The best way to do so would be for the user to be able to use an "endpoint-override" parameter to provide the right endpoint URL. Please  let's discuss it (here](https://github.com/fog/fog-openstack/issues/381). 

Please feel free to re-open  and challenge this if you disagree.

Issue #350, opened 2018-01-09
Failing unit tests locally


timuralp:
Looks like I'm unable to run unit tests locally without failure, as the following tests fail:
```
Fog::Orchestration[:openstack] | stack requests::success#test_0005_#recurse_template_and_file
Fog::Orchestration[:openstack] | stack requests::success#test_0002_#get_file_contents_local_template
```
In both cases, the failure is:
```
Errno::ENOENT: No such file or directory @ rb_sysopen - ///home/travis/build/fog/fog-openstack/test/requests/orchestration/no_recursion.yaml
```

I looked through the changes there, and it looks like it's related to the following change (from 2b32ce8b3d000b20d196601ddbda048d9a463554):
```
diff --git a/lib/fog/orchestration/util/recursive_hot_file_loader.rb b/lib/fog/orchestration/util/recursive_hot_file_loader.rb
index 1de6f15..9361588 100644
--- a/lib/fog/orchestration/util/recursive_hot_file_loader.rb
+++ b/lib/fog/orchestration/util/recursive_hot_file_loader.rb
@@ -123,10 +117,8 @@ module Fog
-            # Don't process file:// outside our base_url. TODO raise an exception here?
-            next if file_outside_base_url?(base_url, str_url)
 
-            file_content = get_content(str_url)
+            file_content = read_uri(str_url)
```

@ioggstream, could you look into this? My understanding is that there is a .yaml template that specifies a URL that conflicts with the base URL used. Previously, this would be skipped over (with a possible TODO for an exception), but now fetching there is an implicit requirement to recreate the `/home/travis/...` build path. I think the solution here could be to mock out the file, verify the path, and return contents from `$FOG_OPENSTACK/test/requests/orchestration/no_recursion.yaml`.

ioggstream:
@timuralp 

The issue you pointed out is not in the library code you referenced but in the test. 

While
 - you can try patching the *test code* (but not the library code) with a clever resolution of $FOG_OPENSTACK/test/... that should work both in travis and on your PC.

I suggest you to either:
 - (not my choiche) create/link/sed /home/travis path  
 - (my choice) developing in a ruby docker container. This allows run tests on different ruby versions in a shot (which I always did).

```

cat >  docker-compose.yml << EOF 
version: '2'

services:

  ruby:
    image: ruby
    entrypoint: tail -f /etc/hosts
    cap_add:
    - SYS_PTRACE
    volumes:
    - .:/code:z
[...]

PR #364, opened 2018-02-12
Broken Orchestration tests: Use relative path

gildub:
Test are broken by the absolute path introduced since https://github.com/fog/fog-openstack/commit/2b32ce8b3d000b20d196601ddbda048d9a463554.

Initially the affected tests failure seemed to be local only but others forks are being impacted too. Therefore the urgent merge.

### DIFF
--- a/lib/fog/orchestration/util/recursive_hot_file_loader.rb
+++ b/lib/fog/orchestration/util/recursive_hot_file_loader.rb
@@ -159,7 +159,6 @@ module Fog
           content = ''
           # open-uri doesn't open "file:///" uris.
           uri_or_filename = uri_or_filename.sub(/^file:/, "")
-
           open(uri_or_filename) { |f| content = f.read }
           content
         end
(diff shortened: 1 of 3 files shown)


timuralp:
@gildub have you considered the discussion in this issue: https://github.com/fog/fog-openstack/issues/350? Specifically, changing these .yaml files actually changes the testing surface. My understanding of the `local_fullpath.yaml` file is that it should test parsing of the full path for the `file://` type paths. I wrote up a PR to set the tests to pending, unless running on Travis, but the real solution would be to either use the docker container that @ioggstream suggested or to mock it out. I don't think changing these files is the right thing to do.

I guess the biggest comment is that there is an open issue for it with a bit of discussion and I wanted to make sure you noticed it.

gildub:
@timuralp, thank you for the follow-up, yes I followed the corresponding issue although I updated it after the PR. 

I understand this issue has been quite annoying and the main blocker, as you have brought it, is breaking the local tests. But not only locally as forked seemed to be impacted too and was blocking new PRs too.

For now, CI and local tests are passing, meanwhile you're right if the test surface is not complete then it should be addressed to.

gildub:
Discussion to be continued on https://github.com/fog/fog-openstack/issues/350

Issue #359, opened 2018-01-29
X-OpenStack-Nova-API-Version header empty value


shoudusse:
My cloud provider only supports v2.0 compute API version.
If i request the version discovery, I have:
```
GET https://ecs.eu-west-0.prod-cloud-ocb.orange-business.com
{u'versions': [{u'id': u'v2.0',
                u'links': [{u'href': u'https://ecs.eu-west-0.prod-cloud-ocb.orange-business.com/v2/',
                            u'rel': u'self'}],
                u'min_version': u'',
                u'status': u'SUPPORTED',
                u'updated': u'2011-01-21T11:33:21Z',
                u'version': u''}]}

```
In this case, If I request the compute API, fog-openstack adds an  `X-OpenStack-Nova-API-Version` HTTP header with an empty value.

Is it the wanted behaviour?

Thanks for your help.

voelzmo:
Hey Sylvain,

is that extra header causing you any issues or is this just something you noticed while inspecting the traffic?

shoudusse:
Hello Marco

Actually, yes, it's causing me an issue as the WAF in front of the Cloud doesn't like this empty Header.
I took the lead opening this issue before asking the WAF team as they will probably answer me that an empty Header is not RFC 7230 compliant.
Regards,

Carthaca:
I would say the latter, checking if `@microversion` is not an empty string in the headers definition.

PR #367, opened 2018-02-15
Test if microversion value not nil and not empty

shoudusse:
These PR fixes https://github.com/fog/fog-openstack/issues/359


voelzmo:
Could you please add a test that asserts the fix works and won't break in the future? Thanks :)

gildub:
Thanks for adding tests 👍

PR #374, opened 2018-02-27
Mock: move code back to /lib/mock to restore feature

gildub:
Moves Mock classes back into `lib/fog/mock` as it can't reside in '/test' only as it needs to be in the main code.


gildub:
@plribeiro3000, per your comment https://github.com/fog/fog-openstack/pull/341#issuecomment-369228400, I'll make the files required. Thanks

PR #375, opened 2018-02-28
Mocks for put_object and public_url

timuralp:
Fixing up a prior PR (https://github.com/fog/fog-openstack/pull/341) that added these mocks.

### DIFF
--- a/lib/fog/storage/openstack.rb
+++ b/lib/fog/storage/openstack.rb
@@ -74,7 +74,15 @@ module Fog
           require_mime_types
           @openstack_api_key = options[:openstack_api_key]
           @openstack_username = options[:openstack_username]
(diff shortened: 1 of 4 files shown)


timuralp:
I squashed two related commits and added a small fix-up for the `public_url` mock. Also, enabled tests for put object after making sure it will return a valid etag for each request.

gildub:
@timuralp, thanks for fixing it.

@giglemad, sorry the discussion about moving built-in Mock was putting a hold on #341, which shouldn't be blocker.

Anyway, this looks good to me.

giglemad:
I could follow the discussion from a distance and was glad to see the discussion and communication expanded to achieve the best level of open source possible. Thanks for this and keep up the good work!

PR #376, opened 2018-03-01
Reverting Mock removal

gildub:
Restores the built-in Mock feature which was killed by Mock classes being moved to `/test/lib/` in https://github.com/fog/fog-openstack/pull/223.

For this to eventually happen a broader work is necessary, this is being discussed [here](https://github.com/fog/fog-core/issues/198)


gildub:
Merging since this is to restore the full Mock feature and doesn't impact the Real classes.

Issue #373, opened 2018-02-27
Excon error on expected status code unmatched


kensoncheung:
Hi there,

I used to use v0.1.19 on both OpenStack Juno and Mitaka version without issues.
Recently I've upgraded fog-openstack to v.0.1.23 for using os-interface related methods.
However, after the upgrade, I found that there are a few API calls returning a correct HTTP status but treated as invalid by Excon when I use it on Mitaka (No issue on Juno).

**Related errors (encountered on Mitaka):**
Create Key Pair: Expected(200) <=> Actual(201 Created)
Delete Key Pair: Expected(202) <=> Actual(204 No Content)

I also reviewed the related API doc, seems like there is status change on related APIs and thus 2 status are both marked as normal response from API doc.
- https://developer.openstack.org/api-ref/compute/#create-or-import-keypair
- Normal response codes: 200, 201
- https://developer.openstack.org/api-ref/compute/#delete-keypair
- Normal response codes: 202, 204

I believe that simply adding the new status codes (201, 204) as accepting status for Excon should be fine. However, I checked the source code for both 0.1.19 and 0.1.23, seems like BOTH are accepting the older status codes (thus Juno will work), but I cannot find the changes of the request methods between these 2 versions. Not sure what cause the difference.  

Please suggest. 

Cheers,
Kenson

PR #378, opened 2018-03-02
Add valid response codes for keypair create/delete.

rdgold94:
Proposed fix for #373 
Change-Id: I028e88e0dfbd8fd7a6b197bda4212b184cc719cc

### DIFF
--- a/lib/fog/compute/openstack/requests/create_key_pair.rb
+++ b/lib/fog/compute/openstack/requests/create_key_pair.rb
@@ -13,7 +13,7 @@ module Fog
 
           request(
             :body    => Fog::JSON.encode(data),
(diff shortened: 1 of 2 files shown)


kjorg50:
@dhague @Ladas - could one of you take a look at this? I think it's a simple fix and our team would like to have this :)

gildub:
@rdgold94, thanks for your help.

rdgold94:
Hi @gildub , just out of curiosity, any idea when the next version of fog-openstack will be out so we know when this fix will be available?

gildub:
@rdgold94, sorry for late response, a new version to be released later this week.... Thanks

PR #379, opened 2018-03-07
Fix error on empty body of File

aufi:
Code at https://github.com/fog/fog-openstack/blob/master/lib/fog/storage/openstack/models/files.rb#L58-L59 can return nil, adapting _body_ method to not fail in this case.


gildub:
@aufi, good catch! Thanks

aufi:
@gildub @Ladas Hi, this PR fixes Fog part of https://bugzilla.redhat.com/show_bug.cgi?id=1551691, could we get a new release?

PR #380, opened 2018-03-07
Fix for https://github.com/theforeman/foreman/pull/5103

pronix:
for not created host id is not exists, but possibility to assign sec.groups should be.
this patch allow return all groups if host not created yet
https://github.com/theforeman/foreman/pull/5103


gildub:
@pronix, thanks

PR #386, opened 2018-03-30
Use autoload

plribeiro3000:
This PR was intended to change some `require` for `autoload` but in the process to fix it i've had to fix other issues i've faced:

* `Planning` was incorrectly under `Openstack` instead of `OpenStack`
* `autoload` for `V2` and `V3` need to be inside its own parent instead of the main file. If we open the class again to define its childs we defeat the purpose of using `autoload` all along.

I've also added a test file to `.gitignore`

cc/ @gildub

### DIFF
--- a/lib/fog/dns/openstack.rb
+++ b/lib/fog/dns/openstack.rb
@@ -1,6 +1,9 @@
 module Fog
   module DNS
     class OpenStack < Fog::Service
+      autoload :V1, File.expand_path('../openstack/v1', __FILE__)
+      autoload :V2, File.expand_path('../openstack/v2', __FILE__)
+
       # Fog::DNS::OpenStack.new() will return a Fog::DNS::OpenStack::V2 or a Fog::DNS::OpenStack::V1,
       # choosing the latest available
       def self.new(args = {})
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ gemfiles/Gemfile-1.9.lock
 .DS_Store
 .idea/
 vendor/bundle
+test/lorem.txt
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -100,7 +100,9 @@ StringLiterals:
   Enabled: false
 StringLiteralsInInterpolation:
   Enabled: false
-TrailingCommaInLiteral:
+TrailingCommaInArrayLiteral:
+  Enabled: false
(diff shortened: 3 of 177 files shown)


lib/fog/dns/openstack.rb
```diff
@@ -1,6 +1,9 @@
 module Fog
   module DNS
     class OpenStack < Fog::Service
+      autoload :V1, File.expand_path('../openstack/v1', __FILE__)
```
gildub: The `__FILE__` is not needed (seems an old practice) and could be replaced with:
```ruby
autoload :V1, 'lib/fog/dns/openstack/v1'
```

And same for all following `autoload`

plribeiro3000: Did not know about this, will check it out. =)

plribeiro3000:
@gildub Rubocop configuration for this repo is 2 years old and never got updated.
I've installed latest version of Rubocop and it could not even parse the configuration. I tried to manually fix one by one but some configurations changed, some don't exist anymore and the project is not consistent with its own styling so i decided to start from scratch and fix the basic ones. Will leave more complex options and options that are kind of personal taste for later.

Will send a PR to my own branch first so i wont pollute this PR.

plribeiro3000:
@gildub Actually i can't do that. The diff is too big. Maybe we can merge this in and i can send small PR's?

gildub:
@plribeiro3000, gotcha re Rubocop issues, thanks for the effort there too! As you said that should be done in a separate PR. 

For now could you please replace the autoload line as suggested so we can merge? 

Thanks

plribeiro3000:
Sure. Will do it now.

plribeiro3000:
@gildub. It breaks the way stuff get loaded. Need to change everything including the tests and fixtures. Maybe we can do it in another PR too?

gildub:
@plribeiro3000, sorry I can't see what it breaks, maybe because of the order libs are loaded.

I've added https://github.com/fog/fog-openstack/pull/388, maybe lets take it from there.

plribeiro3000:
Hmmm. Does the code in your PR work?
It might be that i was doing something wrong then. Will check it tomorrow.

gildub:
@plribeiro3000, could you please rebase? Thanks

gildub:
Replaced by https://github.com/fog/fog-openstack/pull/401

PR #388, opened 2018-04-09
Use autoload with default gem path

gildub:


mrhillsman:
recheck

mrhillsman:
recheck

gildub:
@Ladas, could you please validate? Thanks

PR #391, opened 2018-04-24
Storage: Set cache-control header for a file

dimko:
This PR adds ability to set cache-control header while uploading a file.


timuralp:
Looks good to me! It'd be nice to add a functional test, although I think Swift by default does not store Cache-Control headers and we'd have to check whether the object server is configured to allow the header.

PR #392, opened 2018-04-28
Add configuration for OpenLab

mrhillsman:
reference https://github.com/fog/fog-openstack/issues/385
once this is added we are able to add to OpenLab
configuration to start processing PR payloads

Signed-off-by: Melvin Hillsman <mrhillsman@gmail.com>

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -1,0 +1,35 @@
+- project:
+    check:
+      jobs:
+        - fog-openstack-unittest-test
+        - fog-openstack-unittest-spec
+
+- job:
+    name: fog-openstack-unittest-test
+    parent: init-test
+    description: |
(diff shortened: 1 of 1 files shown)


mrhillsman:
@gildub

.zuul.yaml
```diff
@@ -0,0 +1,35 @@
+- project:
+    check:
+      jobs:
+        - fog-openstack-unit-test
+        - fog-openstack-integration-test
```
gildub: There is no integration tests only unit tests done with VCR which corresponds to spec.

gildub:
@mrhillsman, I suggest to just rename integration unit-test-spec and the unit one unit-test-test or something like that.

PR #401, opened 2018-06-21
Moves autoload to allow reopening

gildub:
Per the idea of https://github.com/fog/fog-openstack/pull/386 this allows service class versions to be autoloaded so they can be reopened properly.


mrhillsman:
recheck

mrhillsman:
recheck

PR #402, opened 2018-06-21
Rake: reduce double definition; Default launch

gildub:
Removes double definitions of tests.
Default now launch both tests


plribeiro3000:
👍

PR #404, opened 2018-06-22
Add additional key servers and keys

mrhillsman:
Jobs are failing due to timeout retrieving keys


mrhillsman:
@gildub please review and merge if ok, this should resolve failing jobs issue due to timeouts retrieving PGP/GPG key

PR #410, opened 2018-08-07
Pin rubocop

tas50:
Make sure we can pull in the current fog-core with fog-openstack.

The rubocop work is just to ensure that doesn't blow up testing.


gildub:
> The rubocop work is just to ensure that doesn't blow up testing.

@tas50, can you please detail ?

tas50:
The next time a rubocop release comes out the repo instances starts failing. Pinning rubocop is a really sane thing to do in a ruby project so you don't have community PRs come in that then need unrelated style cleanup. From time to time you can bump the dep and take on the style cleanup outside of normal work.

gildub:
@tas50, could you please rebase?
Also I would suggest to leave bumping fog-core version to another patch. Because there we're waiting for https://github.com/fog/fog-core/issues/236 to be addressed and resolve a couple of issues. That should help with the test failing. Thanks

gildub:
@tas50, since fog-core 2 is now supported I've changed the title accordingly. Please update you PR so we can move forward.

PR #411, opened 2018-08-07
Don't try to automatically set IDs for new flavors

mansam:
The ID field is optional when creating new flavors in Openstack. If the field is left empty, Openstack will provide a UUID automatically. It's unnecessary for Fog to automatically try to guess the ID, and the ID-guessing code results in strange behavior if a flavor was created outside Fog with a non-integer ID.

https://bugzilla.redhat.com/show_bug.cgi?id=1608230


mansam:
@gildub Can this be backported to v0.1 as well?

PR #413, opened 2018-08-15
New bootstraping

gildub:
In short, it removes as much as possible any "path" tempering.

Authentication returns a token and its catalog along with all services endpoints available. 
When a specific endpoint (among all candidates) is identified, it must be used as it is, without any changes to its path.

There are 2 types of endpoints:
- "version based" such as http://192.0.2.6:8774/v2.1 
- "version less" such as http://192.0.2.6:5000

The difference between those 2 types determines if a prefix is sent along a request.
For example, Nova (https://developer.openstack.org/api-ref/compute/) doesn't prefix any request, beause the endpoint is versionned"

Conversely  when a endpoint is "version less" the API provides the version, for example Neutron (https://developer.openstack.org/api-ref/network/v2/)

The default, from now, is to assume a version less endpoint and therefore no prefix is added.
For service with "versioned" endpoint, the service class, for example Fog::Network::OpenStack, will provide the prefix: 
```ruby
   def default_path_prefix
          'v2.0'
    end
```
Note for reviewers: I'm sorry about this *huge* Pull Request. I've broken down the effort in multiple commits, although the second one is still big. If that causes too much headache we could eventually use a feature branch.

Note: The new authentication, token and catalog approach is under `lib/fog/openstack/auth` and takes about 430 lines of code. The respective unit tests under `unit/auth` are about 845 lines. So altogether we're saving about 200 lines.


liu-sheng:
recheck

aufi:
Tested manually on Queens Packstack installation with some of examples/ scripts and _does not_ work for me.

Example for https://github.com/fog/fog-openstack/blob/master/examples/metric/basics.rb#L4

Works good if AUTH_URL contains endpoint without path. Fails if any path is present, see http://paste.openstack.org/show/728812/

If the I tested it in wrong way, please, let me know, thanks!

gildub:
@aufi, 

Yes the path shouldn't contain 'v3' or 'v2.0' or any 'tokens/auth' or 'such `path. 
The default endpoint provided by Openstack identity should work assuming it's version less.

aufi:
Thanks Gilles for the PR, I think it is a big step forward, I'm going to merge it.

As discussed, error with path in AUTH_URL can be addressed in follow-up PR.

gildub:
Thanks @aufi

PR #414, opened 2018-08-20
Image: Fix attributes to be created as Symbol not String

gildub:


gildub:
@ladas or @aufi, could you please review/merge? Thanks

gildub:
@Ladas, thanks

PR #418, opened 2018-08-28
Allows openstack_auth_url to use version

gildub:
This offers backward compatibility as before the new boostrapping approach.
A deprecation message is added in such case.


gildub:
@aufi, per our discussion, this offers backward compatibility. I've tested with both v2.0 and v3.

PR #419, opened 2018-08-28
Add optional 'name' parameter to backup restore

mansam:
The Cinder API allows providing an optional "name" parameter that will be used when creating a volume to restore the backup to if a volume_id was not passed. This PR adds that parameter to the `restore_backup` request and `restore` method.


gildub:
@mansam, LGTM. Could you please fix the hound/rubocop issue?

mansam:
@gildub thanks for fixing the hounds while I was on PTO, and for merging!

PR #423, opened 2018-09-08
Fix create_temp_url due to missing @host variable

cyrylas:
Class  `Fog::Storage::OpenStack::Mock` is setting `@host` variable, but `Real` version never do that. This breaks `Fog::Storage::OpenStack::Real.create_temp_url()` function, as it depends on `@host` variable.

This pull requests set correct @host variable.


gildub:
@cyrylas, thanks for reporting.

Actually the @host variable is not needed anymore and we could use directly `@openstack_management_uri.host`.

https://github.com/fog/fog-openstack/pull/424 takes care of it.
Let me know if that suits you.

PR #424, opened 2018-09-10
Storage: Use @openstack_management_uri only

gildub:
Uses @openstack_management_uri for scheme, host and port which allow to avoid dedicated instance variables.
Adjusts Mock accordingly.

### DIFF
--- a/lib/fog/storage/openstack.rb
+++ b/lib/fog/storage/openstack.rb
@@ -77,12 +77,8 @@ module Fog
           @openstack_management_url = options[:openstack_management_url] || 'http://example:8774/v2/AUTH_1234'
 
           @openstack_management_uri = URI.parse(@openstack_management_url)
-
-          @host   = @openstack_management_uri.host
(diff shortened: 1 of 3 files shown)


aufi:
Look good to me, merging :+1:

chaadow:
@gildub Hi, thank you for the fix. Can you please update the README to specify that `openstack_auth_url` now does not accept the version, and that you need to use `openstack_identity_api_version: v3.0` instead. ( v2.0 is the default value if i'm not mistaken" thanks again. and if possible release on rubygems 👌

gildub:
@chaadow, You're welcome. 

I'll update the Readme (default is v3 not v2.0) and push a release soon after.

gildub:
https://github.com/fog/fog-openstack/pull/425

gildub:
@chaadow, release 0.2.2 is now available.

chaadow:
@gildub Thank you! I'll update my gem accordingly to support the latest version

Issue #431, opened 2018-09-14
fog-core 2+ support to refactor namespace and clear deprecation messages


gildub:
We now support fog-core 2+ meanwhile we need to re-factor in order to properly align with it.

All deprecation messages such as:
```ruby
[fog][DEPRECATION] The format Fog::Compute::OpenStack is deprecated
[fog][DEPRECATION] Unable to require fog/openstack/compute
```
require to change all Namespace from `Fog::<Service Name>::OpenStack` to `Fog::OpenStack::<Service Name>`

stiller-leser:
Would it be possible to set a less strict fog dependency, i.e. 2.1.0? I know this is an ancient version, but the deprecation warnings introduced into fog-core in september break a lot tests. Same is currently happening with fog-google.

Also thanks for the great work! It is highly appreciated.

ymasson:
thanks a lot @gildub

gildub:
@stiller-leser, @ymasson, version 0.3.1 is now available.

PR #432, opened 2018-09-17
Namespace change to align with fog core 2+

gildub:
- `Fog::<Service>::OpenStack` renamed `Fog::OpenStack::<Service>`
- Files moved from `lib/fog/<service_name>/OpenStack` to `lib/fog/openstack/<service_name>`

https://github.com/fog/fog-openstack/issues/431


aufi:
Looks good to me, but projects relying on fog/openstack will need to update their codebase. Is there a plan on release (possibly keeping v1 and v2 together for some time)?

gildub:
@aufi, we cannot support both v1 and v2 at same time. We have been holding on v1 for a while now and fog-openstack needs to catch-up with fog-core v2. I suppose any project can hold onto a holder version of fog-openstack.

aufi:
I understand that master version will have this PR applied, but it might make a sense to keep e.g. core_v1 branch for possible maintenance releases.

@gildub do you know if the openlab error is related to this PR?

Issue #438, opened 2018-09-24
Image Service doesn't validate supported versions anymore


beyhan:
**Description:**
The fog openstack image service does not check for supported versions anymore. Previously, a `Fog::OpenStack::Errors::ServiceUnavailable` was raised.

**How to reproduce:**
Run the following script with `fog-openstack` version 0.1.19 against an Openstack wich does not support image version 1:
```
require 'fog/openstack'

image = Fog::Image::OpenStack::V1.new({
  :openstack_api_key => “…”,
  :openstack_username => “…”,
  :openstack_auth_url => ‘…’,
  :openstack_project_name => “…”,
  :openstack_project_id => “…”,
  :openstack_user_domain => “…”,
  :openstack_endpoint_type => ‘…’,
})

puts image.images.all
```

The output is
```
/Users/cpi/.rvm/gems/ruby-2.4.4/gems/fog-openstack-0.1.19/lib/fog/openstack.rb:592:in `get_supported_version_path': OpenStack service only supports API versions /v1(\.(0|1))*/ (Fog::OpenStack::Errors::ServiceUnavailable)
	from /Users/cpi/.rvm/gems/ruby-2.4.4/gems/fog-openstack-0.1.19/lib/fog/image/openstack/v1.rb:123:in `set_api_path'
	from /Users/cpi/.rvm/gems/ruby-2.4.4/gems/fog-openstack-0.1.19/lib/fog/image/openstack/v1.rb:115:in `initialize'
	from /Users/cpi/.rvm/gems/ruby-2.4.4/gems/fog-core-1.43.0/lib/fog/core/service.rb:115:in `new'
	from /Users/cpi/.rvm/gems/ruby-2.4.4/gems/fog-core-1.43.0/lib/fog/core/service.rb:115:in `new'
	from /Users/cpi/workspace/fog-examples/lib/image-test.rb:3:in `<main>'

Process finished with exit code 1
```

Then, run the same script with `fog-openstack` version 1.0.0.
The output is
```
/Users/cpi/.rvm/gems/ruby-2.4.4/gems/excon-0.62.0/lib/excon/middlewares/expects.rb:7:in `response_call': Expected([200, 204]) <=> Actual(300 Multiple Choices) (Excon::Error::MultipleChoices)
excon.error.response
[...]

PR #443, opened 2018-09-26
Image::V1: Service type is now imagev1

gildub:
Image service type v1 and v2 need to be differentiated.
Since openstack service type [1] doesn't offer an alias to distinguish between image v1 and v2 and because v1 is depreciate, let's keep the "image" service type to v2 and assign "imagev1" to v1.
While at it, let's also adds "imagev2" as a possible service type for v2.

https://github.com/fog/fog-openstack/issues/438

[1] https://git.openstack.org/cgit/openstack/service-types-authority/

### DIFF
--- a/lib/fog/openstack/image/v1.rb
+++ b/lib/fog/openstack/image/v1.rb
@@ -111,7 +111,7 @@ module Fog
           end
 
           def default_service_type
-            %w[image]
+            %w[imagev1]
           end
         end
       end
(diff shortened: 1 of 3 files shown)


voelzmo:
Is this about image or volume? The code and the commit message seem to be about different services, right?

voelzmo:
@aufi I was talking about the PR description:
> Since openstack service type [1] doesn't offer an alias to distinguish between volume v1 and v2 and because v1 is depreciate, let's keep the "image" service type to v2 and assign "volumev1" to v1.
While at it, let's also adds "volumev2" as a possible service type for v2.

gildub:
@voelzmo, @aufi, sorry I mixed up volume in the description. Fixed!

PR #446, opened 2018-09-27
Auth: Allows an auth_url path to contain version

gildub:
The prefix_path allows to ignore version  ('/v3' or 'v2.0') when the identity endpoint (`auth_url`)
 contains a version in its path.


aufi:
Looks good to me, merging :+1:

Issue #338, opened 2017-11-02
create_temp_url does not support options to allow setting of the disposition


jeffreyguenther:
According to https://docs.openstack.org/swift/latest/middleware.html#tempurl, temporary URLs can have their disposition and filename set as query parameters.

[create_temp_url](https://github.com/fog/fog-openstack/blob/3d5f4ea27a413fd297f230c7c81c9237432bb2bc/lib/fog/storage/openstack/requests/get_object_https_url.rb#L37) currently does not support any options for setting the disposition or filename.

Is there interest in a PR, to add this functionality?

timuralp:
@jeffreyguenther sorry for a very belated response! I think that'd be useful to add. Would you be willing to submit a PR for this?

jeffreyguenther:
I've rotated off the project I would use this on. If I return to the project, I'll open the PR.

PR #452, opened 2018-10-09
Allow custom Content-Disposition filename in #create_temp_url method

vala:
Hi there,

To follow up on #338 I added the support for custom file names in Swift's temp url generation method `Fog::OpenStack::Storage::Real#create_temp_url`.

I found no tests for the `#create_temp_url` method so I did not write any. 

Note that the param is always passed, even when the `:filename` option isn't provided since the Swift service will simply use the last member of the path as the content disposition filename by default, even if the `filename=` param is passed an empty value.

### DIFF
--- a/lib/fog/openstack/storage/requests/get_object_https_url.rb
+++ b/lib/fog/openstack/storage/requests/get_object_https_url.rb
@@ -56,12 +56,14 @@ module Fog
           hmac = Fog::HMAC.new('sha1', @openstack_temp_url_key.to_s)
           sig  = sig_to_hex(hmac.sign(string_to_sign))
 
+          filename = options[:filename]
+
           temp_url_options = {
             :scheme => scheme,
             :host   => host,
             :port   => port,
             :path   => object_path_escaped,
-            :query  => "temp_url_sig=#{sig}&temp_url_expires=#{expires}"
+            :query  => "temp_url_sig=#{sig}&temp_url_expires=#{expires}&filename=#{filename}"
           }
           URI::Generic.build(temp_url_options).to_s
         end


gildub:
@vala, LGTM. Any chance you could add any tests? Obviously we inherit ate disjointed tests but we've got to progressively fill up the blanks.

lib/fog/openstack/storage/requests/get_object_https_url.rb
```diff
@@ -56,12 +56,14 @@ def create_temp_url(container, object, expires, method, options = {})
...
 
+          filename = options[:filename]
+
           temp_url_options = {
             :scheme => scheme,
             :host   => host,
             :port   => port,
             :path   => object_path_escaped,
-            :query  => "temp_url_sig=#{sig}&temp_url_expires=#{expires}"
+            :query  => "temp_url_sig=#{sig}&temp_url_expires=#{expires}&filename=#{filename}"
```
gildub: Filename needs to be filtered out if the parameter is empty.

gildub:
Actually the option need to be filtered if empty, please see inline.

vala:
Why do we need to filter that param out ? 

On my Swift instance, if the filename is empty (`&filename=`) the returned `Content-Disposition` header will not contain any filename, just like when you do not pass the `&filename=` param.

Isn't that your case ?

NikosVlagoidis:
Isn't this fixed [here](https://github.com/fog/fog-openstack/pull/489) ?

ares:
I'm sorry for the delay, there's a conflict now. I think this is no longer relevant as the query can contain multiple keys. Please let me know if I'm missing something, otherwise I think we can now close the PR.

ares:
Closing now, please reopen if I missed anything.

PR #454, opened 2018-10-15
storage: add openstack_storage_url option

elthariel:
In order to support ACL management in my provider (OVH), I need to have users in a different tenant and I need to use another base URL for my storage calls. In the python client it is implemented using the OS_STORAGE_URL.

This is the equivalent option for fog.

I don't really know how to add tests to this, suggestions are highly welcomed

Signed-off-by: Julien 'Lta' BALLET <contact@lta.io>

### DIFF
--- a/lib/fog/openstack/core.rb
+++ b/lib/fog/openstack/core.rb
@@ -16,6 +16,7 @@ module Fog
       attr_reader :openstack_project_id
       attr_reader :openstack_project_domain_id
       attr_reader :openstack_identity_api_version
+      attr_reader :openstack_storage_url
 
       # fallback
       def self.not_found_class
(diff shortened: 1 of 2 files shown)


lib/fog/openstack/core.rb
```diff
@@ -237,6 +230,20 @@ def authenticate
 
         true
       end
+
+      protected
```
gildub: Unless I missed something I believe `private` should be enough.

ares:
@ShamoX seems reasonable to me and should be backwards compatible, thoughts?

ShamoX:
Yes, seems good, but adding a test like @gildub proposed would improve confidence.

@elthariel still need help here to do it (even after @gildub proposition) ?

PR #457, opened 2018-10-23
Swift: Allow 202 for large objects creation/update/copy

gildub:
Adds HTTP return code '202' (Accepted) for Container PUT and COPY requests.

OpenStack object storage API reference guide [1] implicitly mentions return code '202' is expected when dealing with large object for creation, update or copy [2]. 

[1] https://developer.openstack.org/api-ref/object-store/?expanded=copy-object-detail,create-or-replace-object-detail#create-or-update-object-metadata
[2] https://docs.openstack.org/swift/latest/api/large_objects.html

Thanks to @NickLaMuro for reporting this issue.


NickLaMuro:
Awesome, thanks for doing this @gildub !

Also, really appreciate you finding documentation that shows I wasn't completely crazy! 😅👍

gildub:
@NickLaMuro, you're welcome!

Also what is crazy is to have people to maintain such documentation where we could have it automated (Graphql for instance!) ;)

Issue #453, opened 2018-10-15
Auto-retry idempotent requests


NautiluX:
When accessing OpenStack API with idempotent requests, my guess is that should apply to at least all GET or DELETE requests, they should be retried automatically in case of an error.
As an example take the request in [list_ports](https://github.com/fog/fog-openstack/blob/a2b207181ced2136aebae5b870f7872dcb02a17d/lib/fog/openstack/network/requests/list_ports.rb#L6).

It seems like excon already supports options to retry requests as the [README](https://github.com/excon/excon/blob/master/README.md#options) states.

We face failures due to network issues repeatedly, which are not required to fail whole workflows without giving it a second try, hence this would be a very helpful improvement for us.

gildub:
Sounds like a good idea! 

I think we could offer the feature as an option, for instance as a global parameter.
So by default there are no options `{}` and if needed the option could be used on a per request basis or globally with `{:idempotent => true}`.

NautiluX:
That sounds good! However, I think it would be even better for some of the API calls if the option was the default, as they are idempotent in OpenStack. That should for example apply to all the `GET` calls I think.

gildub:
My concern is about the impact on existing deployments and more precisely is on what type of errors the `:idempotent` feature handles. Also what about the others requests PUT/POST, if you have network failures those are likely to fail too, isn't?

NautiluX:
Yes. But in those cases, I think you cannot be sure the request you issued before didn't make it to OpenStack and hence you might, for example, create another disk if you retried the call. That's why those calls should fail. But, calls like GET or DELETE should be easily repeatable as issuing them multiple times should always result in the same state in OpenStack.

Regarding your concerns about existing deployments:
Maybe the global configuration could make it possible to enable idempotent flag for all calls of a given verb? (something like `{:idempotent => ["GET", "DELETE"]}`)
Or the global configuration you mentioned could be an opt-in to all idempotent calls being retried, while the default stays to not retry them.

gildub:
Thanks for the details, it makes sense that GET and DELETE are idempotent by nature vs the modifying verbs.

For the option we could combine both options `{:idempotent => ["GET", "DELETE"]}` and `{:idempotent => true}`. The first determines the VERBS being idempotent and the second activate the feature.

What do you think?

NautiluX:
Sounds good. What would the default be for the list of verbs? Shouldn't it then be something like

```
{
  :idempotent => {
[...]

PR #462, opened 2018-11-02
Auto Retry (Idempotency requests)

gildub:
Auto Retry option for all services

Auto Retry is off by default but can be turned on with `:idempotent` option.
Default values are `:retry_limit => 2` and `:retry_interval => 5` (seconds) 

Only GET and DELETE REST verbs are used for auto retry. More verbs can be added if that make sense.

Addresses https://github.com/fog/fog-openstack/issues/453

### DIFF
--- a/lib/fog/openstack/core.rb
+++ b/lib/fog/openstack/core.rb
@@ -48,6 +48,14 @@ module Fog
       private
 
       def request(params, parse_json = true)
+        if @idempotent && @idempotent_verbs.include?(params[:method])
+          params.merge!(
+            :idempotent     => @idempotent,
+            :retry_limit    => @retry_limit,
+            :retry_interval => @retry_interval
+          )
+        end
+
         retried = false
         begin
           authenticate! if @expires && (@expires - Time.now.utc).to_i < 60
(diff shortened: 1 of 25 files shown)


lib/fog/openstack/core.rb
```diff
@@ -172,6 +180,11 @@ def setup(options)
           instance_variable_set "@#{openstack_param}".to_sym, value
         end
 
+        @idempotent_verbs ||= %w(GET DELETE)
+        @idempotent ||= options[:idempotent].nil? ? false : !!options[:idempotent]
+        @retry_interval ||= options[:retry_interval] ? options[:retry_interval] : 5 # Seconds
+        @retry_limit ||= options[:retry_limit] ? options[:retry_limit] : 2
+
```
aufi: Just a nit, but if I'm not mistaken, 3 lines above could be written in shorter way (also matching to pattern few lines bellow)
```ruby
@idempotent     ||= !!options[:idempotent]
@retry_interval ||= options[:retry_interval] || 5 # Seconds
@retry_limit    ||= options[:retry_limit] || 2
```

gildub: That's neat.

Absolutely, I'lll looking into adding some tests

gildub: @aufi, sorry didn't get the bandwidth to do it.

aufi:
@gildub Hi Gilles, I know it is late, but I'm happy merging this PR when the https://github.com/fog/fog-openstack/pull/462/conflicts gets resolved.

gildub:
@aufi, Hi Markek, sorry for slow response. I've addressed conflict. Thanks!

PatrickFranken:
@aufi any chance to merge this into the master? The Swift service used by my company is pretty unreliable and this should solve that issue.

ares:
@ShamoX this seems as a good and safe addition and ready to get in. Of course tests are missing, what is your preference here?

ShamoX:
Yes, tests are missing. I don't really see how we can tests this except by implementing a complexe request/response scenario, which seems not easy to implement with VCRs (maybe with a sort of interaction with doubles - but seems really complicated).

Anyway, I can review it more precisely when I have more time and see if I can catch any statically analysable bad behaviours. But I doubt it.

We should at least test new configuration variables to see if they are used.

Issue #469, opened 2019-01-04
planning.rb: Don't let binding.pry into production, please ^^


ShamoX:
I got stuck several times in different contextes on a ligne of `planning.rb`. I made a little search of binding.pry and found it only there. Is there a reason why you let this statement inside published code ?
```shell
$ grep binding.pry -rnB 2 lib
lib/fog/openstack/planning/planning.rb-122-    def self.new(attributes)
lib/fog/openstack/planning/planning.rb-123-      require 'pry-byebug'
lib/fog/openstack/planning/planning.rb:124:      binding.pry
```

I'll prepare a pull request.

gildub:
@ShamoX, this was just a debug stanza left behind. Thanks

PR #470, opened 2019-01-04
Remove binding.pry statement and related.

ShamoX:
Fix for #469

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: ruby
 sudo: false
 before_install:
-  - gem install bundler --version '~> 1.6'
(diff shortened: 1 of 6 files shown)


ShamoX:
Ok, travis builds seems to be broken. I try to fix them

ShamoX:
Maybe I misunderstood the presence of the binding.pry.
If you want to prevent people to use this, maybe we can add a deprecation message and/or raise an exception and direct the developper to `Fog::OpenStack.service_klass(:Compute)` for example ?

aufi:
Thanks for the fix, merging :+1:

PR #473, opened 2019-01-08
Fix typo

alexander-demichev:
Fix typo in method call

@aufi @gildub


aufi:
@alexander-demichev Don't we missing question mark in the end of method?

PR #482, opened 2019-01-17
Add multiattach parameter to volume creation

mansam:
https://bugzilla.redhat.com/show_bug.cgi?id=1660896

@gildub What if anything do I need to do here to support older versions of Cinder that don't know about this parameter?

https://docs.openstack.org/cinder/latest/admin/blockstorage-volume-multiattach.html


gildub:
@mansam, do you mean "How to make volume v2 to support v3"? If yes then version v3 is likely needed.

gildub:
@mansam, I just realized V2 does has multiattach option for create volume. My bad! 
Well V3 is still needed but that's another story. 
Merging now!

PR #483, opened 2019-01-18
Dell ECS requires the Accept header to be set to application/json, too.

boof:
Validated that this now works with our Dell ECS instance. It requires the Accept header to be set to return JSON. If not It'll return XML which will result in a ParseError.


mrhillsman:
recheck

boof:
I think the VCR casettes need to be regenerated.

jablan:
@boof I opened https://github.com/fog/fog-openstack/pull/495

Issue #487, opened 2019-01-31
Problem with number values in connection_params


vadimstroganov:
Hello!

I have a problem with the connection_params, if I specify values consisting only of numbers.

for-core call `coerce_options` for connection_params.

`coerce_options` suggests it takes a Hash and converts all the values to underlying Ruby types. So a username that is only numbers is converted to a number.

After the conversion, the string "12345" is converted to the number 12345. And after that, a request for authentication of OpenStack API is sent.

In this case, OpenStack API in the response body is an empty string.

Then an error occurs. (can't decode JSON; response.body = "")

https://github.com/fog/fog-openstack/blob/master/lib/fog/openstack/auth/token/v3.rb#L58

I already wrote about this here https://github.com/fog/fog-core/issues/247

Probably the best way would be to wrap the parameters after call coerce_options into a string?

vadimstroganov:
I created PR https://github.com/fog/fog-openstack/pull/488

cyrylas:
Is that related to #426? I think I have already fixed that.

vadimstroganov:
No, https://github.com/fog/fog-openstack/pull/426 about `tenantId` and `tenantName` keys.
My problem arises with `openstack_` keys.

vadimstroganov:
In https://github.com/fog/fog-openstack/pull/488 for the keys starting with `openstack_` I call a conversion to string.

PR #488, opened 2019-01-31
Wrap OpenStack API credentials to string

vadimstroganov:
Fix it. https://github.com/fog/fog-openstack/issues/487

### DIFF
--- a/lib/fog/openstack/storage.rb
+++ b/lib/fog/openstack/storage.rb
@@ -113,6 +113,7 @@ module Fog
         end
 
         def initialize(options = {})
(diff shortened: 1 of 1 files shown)


gildub:
@vadimstroganov, 

Thanks for looking into this. I think we could benefit from it across the board by using it at the super level:
https://github.com/fog/fog-openstack/blob/master/lib/fog/openstack/core.rb#L42

gildub:
@vadimstroganov, I looked at this but I don't think this is the right answer for it. We cannot change the options on the fly without having a risk of side effects. If there is an issue with Fog::JSON then we should address it there.

PR #489, opened 2019-02-05
Allow custom filename through Content-Disposition header

cyrylas:
Allow to set filename for downloaded file.
Basically 
fog-openstack_before
becomes
fog-openstack_after

### DIFF
--- a/lib/fog/openstack/storage/requests/get_object_https_url.rb
+++ b/lib/fog/openstack/storage/requests/get_object_https_url.rb
@@ -61,7 +61,7 @@ module Fog
             temp_url_sig: sig_to_hex(hmac.sign(string_to_sign)),
             temp_url_expires: expires
           }
-          query[:filename] = options[:filename] unless options[:filename].nil?
+          query[:filename] = options[:filename] if options[:filename]
 
           temp_url_options = {
             :scheme => scheme,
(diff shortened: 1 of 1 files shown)


lib/fog/openstack/storage/requests/get_object_https_url.rb
```diff
@@ -54,14 +56,19 @@ def create_temp_url(container, object, expires, method, options = {})
           string_to_sign = "#{method}\n#{expires}\n#{object_path_unescaped}"
 
           hmac = Fog::HMAC.new('sha1', @openstack_temp_url_key.to_s)
-          sig  = sig_to_hex(hmac.sign(string_to_sign))
+
+          query = {
+            temp_url_sig: sig_to_hex(hmac.sign(string_to_sign)),
+            temp_url_expires: expires
+          }
+          query[:filename] = options[:filename] unless options[:filename].nil?
```
aufi: Thanks for the PR! Just a nit, it might be easier to read ```query[:filename] = options[:filename] if options[:filename]```

cyrylas: Sure, code updated. Thanks :)

gildub:
@cyrylas, that looks better, thanks!

PR #491, opened 2019-03-08
Merge new master branch to old fix-rubocop one

ShamoX:


ShamoX:
OK, there is still remaining cops offended. But They are much touchy. I don't know when I'll be able to do then and if we can fix them.

gildub:
@ShamoX, that sounds good meanwhile I believe you have to start from latest commit because you PR is containing lots of already committed ones. Thanks

ShamoX:
@gildub I'll work again on this when #521 will be merged since it also alter Rubocop behaviour in a way that seems to me with a higher priority. What do you think ?

ShamoX:
Hello,

Should we merge this and start working on ruby 2.7 and 3.x support (as asked by #522 ) ?

PR #493, opened 2019-06-26
Ensure User's Password is always String

aufi:
When OpenStack User's password was a Number, fog-core coerced it to Numeric type, which caused
HTTP 400 failure leading to print request malformed error message including the password.

Value's coercing function from fog-core is a feature and it should not be changed propably,
so adding String cast directly to OpenStack provider.

Related to https://bugzilla.redhat.com/show_bug.cgi?id=1723180

### DIFF
--- a/lib/fog/openstack/core.rb
+++ b/lib/fog/openstack/core.rb
@@ -172,6 +172,9 @@ module Fog
           instance_variable_set "@#{openstack_param}".to_sym, value
         end
 
+        # Ensure OpenStack User's Password is always a String
+        @openstack_api_key = @openstack_api_key.to_s if @openstack_api_key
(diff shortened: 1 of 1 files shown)


aufi:
Travis fail for Ruby 2.5 seems not related to me (other versions pass).

aufi:
Added Travis fix with bundler https://github.com/fog/fog-openstack/pull/498

aufi:
Travis is :green_apple:

aufi:
Thanks for merge :+1: 

Could we have a new release with this change? @Ladas @gildub @geemus

geemus:
Looks like @ladas pushed a release.

PR #495, opened 2019-07-26
Request JSON from auth server

jablan:
Fixes #483


jablan:
seems there is some problem with bundler versions, the CI failure is not related with the change

PR #497, opened 2019-08-07
[WIP] Update Orchestration Stack find by ID method

aufi:
If searching in all Stacks including nested ones is needed, it is necessary
use find Stack by ID as the only argument.

Heat API supports this call [1]. Updating current find_by_id method to
use this API call and making it not-deprecated.

Related to https://github.com/ManageIQ/manageiq-providers-openstack/pull/487

[1] https://docs.openstack.org/api-ref/orchestration/v1/?expanded=find-stack-detail,show-stack-details-detail#find-stack

### DIFF
--- a/lib/fog/openstack/orchestration.rb
+++ b/lib/fog/openstack/orchestration.rb
@@ -47,6 +47,7 @@ module Fog
       request :show_resource_metadata
       request :show_resource_schema
       request :show_resource_template
+      request :show_stack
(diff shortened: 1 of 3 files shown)


andyvesel:
@aufi Reopened PR https://github.com/fog/fog-openstack/pull/505/commits/4130c6921d3e8d2ef834132b0514b44dda8e36e6

PR #498, opened 2019-08-27
Remove bundler pessimistic 1 version

aufi:
Bundler ~> 1 dependency failed on certain Ruby version. Removing version requirement.


aufi:
@ladas @gildub Could you review? Thanks

Issue #503, opened 2019-10-08
Invalid input for field 'scope/project/name': *** is not of type 'string'


tonymadbrain:
Hello!
After provider upgraded OpenStack to Queens we have error when try to auth:

`Actual(400 Bad Request) (Excon::Error::BadRequest)
excon.error.response
  :body          => "{\"error\": {\"message\": \"Invalid input for field 'scope/project/name': 123 is not of type 'string'\\n\\nFailed validating 'type' in schema['properties']['scope']['properties']['project']['properties']['name']:\\n    {'type': 'string'}\\n\\nOn instance['scope']['project']['name']:\\n    23446\", \"code\": 400, \"title\": \"Bad Request\"}}"
  :cookies       => [
  ]
  :headers       => {
    "Content-Length"         => "326"
    "Content-Type"           => "application/json"
    "Date"                   => "Tue, 08 Oct 2019 11:23:47 GMT"
    "Server"                 => "Apache/2.4.18 (Ubuntu)"
    "Vary"                   => "X-Auth-Token"
    "X-Distribution"         => "Ubuntu"
    "x-openstack-request-id" => "req-4c4f7e09-d41b-4ad7-85cf-2496cb819b01"
  }
  :host          => "auth.some.openstack.provider.com"
  :local_address => "*.*.*.*"
  :local_port    => 59461
  :path          => "/v3/auth/tokens"
  :port          => 5000
  :reason_phrase => "Bad Request"
  :remote_ip     => "*.*.*.*"
  :status        => 400
  :status_line   => "HTTP/1.1 400 Bad Request\r\n"`

I did small fix (pr attached). Is this ok?

PR #504, opened 2019-10-08
#503 fixed type for auth

tonymadbrain:


tonymadbrain:
#503

aufi:
Thanks for the PR, the change looks reasonable :+1:  However the ```lib/fog/openstack/auth/name.rb``` file looks as a technical debt to me. It would be great address the file refactoring later.

PR #505, opened 2019-10-14
Update Orchestration Stack find by ID method

andyvesel:
If searching in all Stacks including nested ones is needed, it is necessary
use find Stack by ID as the only argument.

Heat API supports this call [1]. Updating current find_by_id method to
use this API call and making it not-deprecated.

Related to https://github.com/ManageIQ/manageiq-providers-openstack/pull/487

[1] https://docs.openstack.org/api-ref/orchestration/v1/?expanded=find-stack-detail,show-stack-details-detail#find-stack


aufi:
Looks good to me :+1:

Issue #492, opened 2019-06-11
broken auth_url


rokroskar:
It seems that the examples/suggested configs where the `openstack_auth_path` is set to `<host-url>/v3/auth/tokens` don't work with the current version of fog. Using GitLab 11.9.12 for example, the auth url incorrectly becomes `<host-url>/v3/auth/tokens/auth/tokens`. GitLab appears to be using fog `1.0.8`: https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/21788/diffs#21fb5836b499a2be648386aac055d2e069160d6c_247_246

nsteinmetz:
Hi @rokroskar,

I have the same issue, did you find a way to get it working ?

Thanks!

I opened also https://gitlab.com/gitlab-org/gitlab/issues/208825

rokroskar:
Hi @nsteinmetz, the solution is just to strip out `/auth/tokens` at the end of `openstack_auth_url`. 

Hope that helps!

nsteinmetz:
Seems the bug changed a bit since yours in fact.

Seems now the `/v2/tokens/` is added automatically so I can't use a V3 endpoint whereas OVH (my openstack provider) will shut down V2 endpoint by end of the month...

rokroskar:
Surely they didn't downgrade from `v3` api to `v2`? A silly thought but maybe worth trying - you have a `/` after `v3` in your example in the gitlab issue - did you try it without? I.e. `https://auth.cloud.ovh.net/v3` instead of `https://auth.cloud.ovh.net/v3/`?

nsteinmetz:
@rokroskar I tried indeed but forgot to mention it - and it fails too.

It leads to `path          => "/v3/v2.0/tokens"`

axi:
Hi, @nsteinmetz did you managed to find a solution ? I'm having the same problem.
Ovh replanned the v2 shutdown to 23 june 2020...
Thanks

nsteinmetz:
Hi @axi 

I switched to S3 endpoints on OVH instead of swift for the time being.

The issue on gitlab side is in backlog but no progress for a month.

ShamoX:
To use v3 authentication endpoint on OVH :

- auth_url: https://auth.cloud.ovh.net/
- username
- api_key
- project_id

[...]

PR #513, opened 2020-07-20
Fix URLs to use v3 when only v2 endpoint is listed in the catalog

achevalet:
It covers the cases where we can authenticate with keystone v3 but the endpoint returned in the catalog is v2
see https://github.com/fog/fog-openstack/issues/492

You guys can probably do something much cleaner as I have poor ruby skills, but it fixes my issue and the tests passed so I'm proposing.. :-)


pdion891:
tested on an OpenStack not exposing a endpoint for v3 and this fix work fine.

ShamoX:
Just have a little problem with that, it's missing tests implementation of the corrected bug.

I can believe @pdion891 on word but it seems a little risky for the others.

Implemented tests are passing anyway.

ares:
A test would certainly be good, @achevalet please let us know if you plan to add one.

pdion891:
> Just have a little problem with that, it's missing tests implementation of the corrected bug.
> 
> I can believe @pdion891 on word but it seems a little risky for the others.
> 
> Implemented tests are passing anyway.

For sure I've only validated the use case that was valid on my side where the /v3 is not exposed.

achevalet:
Quite busy, I would be grateful if someone can contribute here

PR #517, opened 2021-07-27
Remove ipaddress dependency

ekohl:
In 118a99282e1225db8735f5547280788ede97a09e this was added, but I can't find any place that it would be used.


ShamoX:
This seems true to me too.

ares (approved):
Thanks @ekohl, merging!

PR #523, opened 2022-03-13
setup shared actions/workflows

geemus:
Greetings. I've been trying to work through the repos and get them setup on shared github actions config for consistency and ease of maintenance. This should make those changes here, which include a couple of things of potential note:

1. It changes from using specific test actions within rake to the default rake action (which includes the specific ones, but also the mock ones, which weren't previously included, fwiw).
2. I believe it will slightly change the set of utilized ruby versions. I'm going to try to maintain the shared workflows to use whatever the currently supported versions of Ruby are, so hopefully this will help us stay up to date on that.
3. It adds the stale bot. I have found this to be a useful way to keep things a bit more under control with issues, but we can discuss if you disagree (I've found this can be a bit controversial, depending on who you ask). If nothing else, I think it helps ensure that we periodically give at least a brief look at things instead of letting them go forgotten.
4. It adds a dependabot config. This should help us keep track of any security issues and it also looks for new actions versions, which I think should help us keep the shared workflows in sync as well.

Let me know if you have any questions/concerns. My hope is that it should be largely a drop-in replacement for the most part while also helping me to make things a bit more consistent and maintainable across the suite of libraries.


ShamoX:
Hello,

I thought I made them work ^^

I go take a look.

ShamoX:
Ok, then I understood a little more how the cassette are supposed to work. It seems that the devstack install is missing the identity catalog. I have to dig about that.

It should be better to integrate a script to install on a fresh install of Ubuntu 20.04 somewhere in the doc. I will start from what is explained [here](https://docs.openstack.org/devstack/latest/).

Anyway your changes here seems ok to me, I will first create an other PR to correct the fixtures/cassettes and improve the documentation about setting up a devstack.

ShamoX:
Hello,

I'm working now to be able to start the devstack on a reproductible environment (docker) with the docker-compose scripts.

This will be a first response to #111. My work is visible here [Shamox/fog-openstack/ShamoX/doc-and-tests](https://github.com/ShamoX/fog-openstacktree/ShamoX/doc-and-tests).

I'll make a PR when it's ready (working + doc).

PR #527, opened 2022-10-08
Add partial support for ruby 3 / 3.1

### LINKED ISSUE #522: Ruby 3.0 support
Recently I bumped ruby to 3.0 and received this error:
```text
NoMethodError: undefined method `encode' for URI:Module
  from fog-openstack (1.0.11) lib/fog/openstack/storage/requests/delete_multiple_objects.rb:48:in `block in delete_multiple_objects'
  from fog-openstack (1.0.11) lib/fog/openstack/storage/requests/delete_multiple_objects.rb:46:in `map'
  from fog-openstack (1.0.11) lib/fog/openstack/storage/requests/delete_multiple_objects.rb:46:in `delete_multiple_objects'
  from activestorage-openstack (1.5.2) lib/active_storage/service/open_stack_service.rb:78:in `block in delete_prefixed'
  [...]
```

`URI.escape` (`URI.encode` is an alias) was already obsolete in ruby 2.7 and removed in 3.0.

Are there chances this will be fixed?

chaadow:
This basically replaces all `URI.encode` occurences with `CGI.escape`

As the maintainer of [activestorage-openstack](https://github.com/chaadow/activestorage-openstack).   this has made my [test suite](https://github.com/chaadow/activestorage-openstack/actions/runs/3211049861) pass both on ruby `3.0` and `3.1` so the change should be enough for the storage component

It would be much appreciated if a release is made with this PR merged 🙏🏼 

Fixes #522

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /.bundle/
 /.yardoc
 /Gemfile.lock
+/tags
 /_yardoc/
 /coverage/
 /doc/
--- a/lib/fog/openstack/compute/requests/delete_service.rb
+++ b/lib/fog/openstack/compute/requests/delete_service.rb
@@ -4,7 +4,7 @@ module Fog
       class Real
         def delete_service(uuid, optional_params = nil)
           # Encode all params
-          optional_params = optional_params.each { |k, v| optional_params[k] = URI.encode(v) } if optional_params
+          optional_params = optional_params.each { |k, v| optional_params[k] = CGI.escape(v) } if optional_params
 
           request(
             :expects => [202, 204],
--- a/lib/fog/openstack/compute/requests/disable_service.rb
+++ b/lib/fog/openstack/compute/requests/disable_service.rb
@@ -6,7 +6,7 @@ module Fog
           data = {"host" => host, "binary" => binary}
 
           # Encode all params
-          optional_params = optional_params.each { |k, v| optional_params[k] = URI.encode(v) } if optional_params
+          optional_params = optional_params.each { |k, v| optional_params[k] = CGI.escape(v) } if optional_params
 
           request(
             :body    => Fog::JSON.encode(data),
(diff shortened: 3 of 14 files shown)


maciejpankanin:
I don't think `CGI.escape` is a good choice here since we need URL escaping. See https://stackoverflow.com/a/13059657/7998769.

I think using `addressable` gem would be safer.

geemus:
I think the bigger question is whether any of these should keep `/` or not (ie it looks like some names might include this, which probably shouldn't be escaped but CGI would escape it). Can anyone speak more authoritatively on whether we should be concerned about this?

tas50:
This correctly identifies all thirteen `URI.encode` call sites, which is the hard part — thanks for finding them.

One thing worth flagging on the replacement, though: `CGI.escape` encodes a space as `+`, and these are all URL **paths** rather than query strings. In a path a `+` is a literal plus, not a space:

```ruby
CGI.escape("sp ace")             # => "sp+ace"
Fog::OpenStack.escape("sp ace")  # => "sp%20ace"
```

So `get_action("my action")` would request `actions/my+action` and look up a resource literally named `my+action`.

`delete_multiple_objects` is affected more seriously. It escapes the already-joined string:

```ruby
object_name = container ? "#{container}/#{name}" : name
```

`CGI.escape` percent-encodes that separator, so the body becomes `container%2Fobject` instead of `container/object`, which breaks Swift bulk delete for every object.

The gem already has a helper for exactly this — `Fog::OpenStack.escape`, commented *"CGI.escape, but without special treatment on spaces"* — and the storage requests have been using it all along. It takes an `extra_exclude_chars` argument for the separator case, which is how `Storage::Files#get_url` handles the same problem:

```ruby
Fog::OpenStack.escape(object_name, '/')
```

I have opened #554 taking that approach, with specs covering the space and separator cases. It is built on the call sites you identified here — happy to close mine if you would rather update this one instead.

Issue #510, opened 2020-04-14
Support authenticating with application credentials


Lasering:
Application credentials provide a way to delegate a user’s authorization to an application without sharing the user’s password authentication. This is very useful to test Chef cookbooks using kitchen.

https://docs.openstack.org/api-ref/identity/v3/index.html?expanded=authenticating-with-an-application-credential-detail#authenticating-with-an-application-credential

PR #533, opened 2024-03-18
Add support for using application credentials for authentication

scornelissen85:
I've used this code for using application credentials for authentication in Openstack. I think it would be valuable if this could be added in the fog-openstack project.

This will fix issue #510

This can be used as alternative next to it's original credentials options. To use application credentials I've set the following parameters:

```
compute_client = Fog::Compute.new  :openstack_auth_url => ENV['OS_AUTH_UR'],
                                    :openstack_application_credential_secret => ENV['OS_APPLICATION_CREDENTIAL_SECRET'],
                                    :openstack_application_credential_id => ENV['OS_APPLICATION_CREDENTIAL_ID']
```

CC @dhague (sorry if I should cc another member)

### DIFF
--- a/lib/fog/openstack/auth/token/v3.rb
+++ b/lib/fog/openstack/auth/token/v3.rb
@@ -18,6 +18,11 @@ module Fog
                            'methods' => ['token'],
                            'token'   => {'id' => @token}
                          }
+                       elsif @application_credential
+                         {
+                           'methods' => ['application_credential'],
+                           'application_credential' => @application_credential
+                         }
                        else
                          {
                            'methods'  => ['password'],
@@ -50,6 +55,7 @@ module Fog
           end
 
           def scope
+            return false if @application_credential
             return @project.identity if @project
             return @domain.identity if @domain
           end
@@ -96,6 +102,11 @@ module Fog
 
             if auth[:openstack_auth_token]
               @token = auth[:openstack_auth_token]
+            elsif auth[:openstack_application_credential_id] and auth[:openstack_application_credential_secret]
+              @application_credential = {
+                :id => auth[:openstack_application_credential_id],
+                :secret => auth[:openstack_application_credential_secret],
+              }
             else
               @user = Fog::OpenStack::Auth::User.new(auth[:openstack_userid], auth[:openstack_username])
               @user.password = auth[:openstack_api_key]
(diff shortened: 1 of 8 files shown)


lib/fog/openstack/auth/token/v3.rb
```diff
@@ -50,6 +55,7 @@ def path
           end
 
           def scope
+            return false if @application_credential
```
geemus, line 58: It feels to me like `nil` might be a bit more appropriate than false here. It should still return falsey in the conditional, but just conceptually a nil scope makes more sense to me than a false scope. Does that make sense?

geemus:
Looks good overall.

Unfortunately the test suite has been busted for a bit here, other maintainers haven't really been active, and I don't have an active openstack setup to test against either.

Have you been able to test this out successly?

scornelissen85:
> Looks good overall.
> 
> Unfortunately the test suite has been busted for a bit here, other maintainers haven't really been active, and I don't have an active openstack setup to test against either.
> 
> Have you been able to test this out successly?

Yes, I'm using the branch with these changes with success and only use application credentials to log in in Openstack. I'm using this branch currently from beaker-openstack where the Compute and the Network class is used directly. I'm not sure if something is missing as I'm not using anything other than those classes.

Furthermore, I'm not an expert about the communication with Openstack. This change works for me, but I can't guarantee it works for other cases.

lib/fog/openstack/auth/token/v3.rb
scornelissen85: Looks fine to me to change it to `nil`. I will test if it then still works.

geemus:
Thanks for the context, that makes sense to me. The change to nil would be great and then I think we can bring this in. You can only speak for your own experience, but from what I see it shouldn't break anything that is currently working for others either way.

scornelissen85:
I have changes `scope` to return `nil` when using application credentials and it seems to work.

geemus:
Great, thanks for that.

geemus (approved):

geemus:
@scornelissen85 did you need a release with this change?

scornelissen85:
@geemus If you could make a release I can then make a PR for beaker-openstack to use the application credentials.

Issue #535, opened 2024-04-09
undefined method `=~' for 3:Integer


badde-media:
I have a rather fresh install of activestorage-openstack, which seems to build upon fog-openstack.

I've tried to upload a test jpg file to my Infomaniak s3 comp object storage, but I get this error:

```
Loading development environment (Rails 7.1.3.2)
irb(main):001* blob = ActiveStorage::Blob.create_and_upload!(
irb(main):002*   io: File.open("./public/sleep.jpg", 'rb'),
irb(main):003*   filename: "sleep.jpg",
irb(main):004*   content_type: "image/jpg"
irb(main):005> )
  TRANSACTION (0.0ms)  begin transaction
  ActiveStorage::Blob Create (0.3ms)  INSERT INTO "active_storage_blobs" ("key", "filename", "content_type", "metadata", "service_name", "byte_size", "checksum", "created_at") VALUES (?, ?, ?, ?, ?, ?, ?, ?) RETURNING "id"  [["key", "irdd44jxhg8zzh9ado2ogy61kzdj"], ["filename", "sleep.jpg"], ["content_type", "image/jpeg"], ["metadata", "{\"identified\":true}"], ["service_name", "infomaniak"], ["byte_size", 526621], ["checksum", "H3N/tmO1/TcKfCdnEyIOOQ=="], ["created_at", "2024-04-09 19:22:37.714505"]]
  TRANSACTION (6.6ms)  commit transaction
  OpenStack Storage (9.7ms) Uploaded file to key: irdd44jxhg8zzh9ado2ogy61kzdj (checksum: H3N/tmO1/TcKfCdnEyIOOQ==)
/home/un/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/bundler/gems/fog-openstack-794aa8b36a06/lib/fog/openstack/auth/token.rb:17:in `build': undefined method `=~' for 3:Integer (NoMethodError)

It works when I use   config.active_storage.service = :local but that's ont the intention. 

Copilot keeps telling me that I should fork your gem, rewrite it and try again.

I have already tried @chaadow's version but that has the same error

Please let me know if I can help

geemus:
It looks like it's receiving an integer for the api version, while expecting a string (and then erroring because it's trying to compare a regex to an integer instead of a string) here: https://github.com/fog/fog-openstack/blob/master/lib/fog/openstack/auth/token.rb#L17

[...]

PR #536, opened 2024-04-10
Update token.rb

badde-media:
Trying to fix this error (https://github.com/fog/fog-openstack/issues/535) by changing the input to_s before trying to match it with the regex

### DIFF
--- a/lib/fog/openstack/auth/token.rb
+++ b/lib/fog/openstack/auth/token.rb
@@ -14,7 +14,7 @@ module Fog
         class URLError < RuntimeError; end
 
         def self.build(auth, options)
-          if auth[:openstack_identity_api_version] =~ /(v)*2(\.0)*/i ||
+          if auth[:openstack_identity_api_version].to_s =~ /(v)*2(\.0)*/i ||
              auth[:openstack_tenant_id] || auth[:openstack_tenant]
             Fog::OpenStack::Auth::Token::V2.new(auth, options)
           else


badde-media:
Using my fixed version, this bug is actually gone, but I still get an auth error

geemus:
@badde-media progress at least, as you said. Could you share what your credentials look like (please be sure to sanitize them by putting in placeholder values for any secrets) and the details of the new error you are now seeing? Thanks.

badde-media:
Hello again, sorry for the late reply. I was getting crazy with some other problems this app had. 

This is the storage.yml I've used that finally worked with Infomaniak S3 Object Storage:
```
infomaniak:
   service: S3
   access_key_id: <%= Rails.application.credentials.dig(:infomaniak, :access_key_id) %>
   secret_access_key: <%= Rails.application.credentials.dig(:infomaniak, :secret_access_key) %>
   region: "us-east-1"
   bucket: "myBucketName"
   endpoint: "https://s3.pub1.infomaniak.cloud"
   public: true
```   
using the version of this gem that I had patched. So I didn't even have to enter the version.
Here's another version that also could connect, as far as I can remember:
```
 openstack:
     auth:
       auth_url: https://api.pub1.infomaniak.cloud/identity
       username: "username"
       project_id: project_id
       project_name: "project_name"
       user_domain_name: "Default"
     region_name: "dc3-a"
     interface: "public"
     identity_api_version: 3
```
In this case, the identity_api_version only worked when entered like this, just as an Integer. no "v3" or something like that.

BUT one of the aforementioned problems is that Infomaniak S3 DOES NOT SUPPORT CORS, wich renders this service useless for live-streaming via HTTPS directly from the object storage, and also for direct uploads as ActiveStorage offers it. I guess that Infomaniak want to sell their more expensive audio- and video-streaming servers for this purpose. 
This has nothing to do with fog-openstack, but I'm mentioning it just in case that others run into the same problems that I have had, and where you can hardly find any information online. Because of this CORS error, we're now switching most probably to DigitalOcean. This is my DO storage.yml entry that also works using my patched version:
```
digitalocean:
  service: S3
  endpoint: https://fra1.digitaloceanspaces.com
  access_key_id: <%= Rails.application.credentials.dig(:digitalocean, :access_key) %>
  secret_access_key: <%= Rails.application.credentials.dig(:digitalocean, :secret) %>
  bucket: myBucketName
  region: unused
```
(nb: do not enter the bucket name in the endpoint url)
This also just for completeness.

I guess we can close this commit, no?

geemus:
Thanks for the update, sure I will close for now.

PR #543, opened 2025-03-14
Application credentials support

andrejcermak:
Hi, I added the application credential support that was missing in this library.

### DIFF
--- a/lib/fog/openstack/identity/v3/requests/create_application_credentials.rb
+++ b/lib/fog/openstack/identity/v3/requests/create_application_credentials.rb
@@ -1,0 +1,23 @@
+module Fog
+    module OpenStack
+      class Identity
+        class V3
+          class Real
+            def create_application_credentials(credential = {})
+              user_id = credential.delete('user_id') || credential.delete(:user_id)
+              puts Fog::JSON.encode(:application_credential => credential)
+              request(
+                :expects => [201],
+                :method  => 'POST',
+                :path    => "/users/#{user_id}/application_credentials",
+                :body    => Fog::JSON.encode(:application_credential => credential)
+              )
+            end
+          end
+  
+          class Mock
+          end
+        end
+      end
+    end
+  end
--- a/lib/fog/openstack/identity/v3/requests/get_application_credentials.rb
+++ b/lib/fog/openstack/identity/v3/requests/get_application_credentials.rb
@@ -3,20 +3,22 @@ module Fog
       class Identity
         class V3
           class Real
-            def list_application_credentials(user_id)
+            def get_application_credentials(id, user_id)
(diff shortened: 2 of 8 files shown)


andrejcermak:
@geemus any chance of getting this merged?

lib/fog/openstack/identity/v3/requests/create_application_credentials.rb
```diff
@@ -0,0 +1,23 @@
+module Fog
+    module OpenStack
+      class Identity
+        class V3
+          class Real
+            def create_application_credentials(credential = {})
+              user_id = credential.delete('user_id') || credential.delete(:user_id)
+              puts Fog::JSON.encode(:application_credential => credential)
```
geemus, line 8: I suspect this was unintentionally left behind from testing/debugging.
```suggestion
```

lib/fog/openstack/identity/v3/requests/create_application_credentials.rb
```diff
@@ -0,0 +1,23 @@
...
+              request(
+                :expects => [201],
+                :method  => 'POST',
+                :path    => "users/#{user_id}/application_credentials",
+                :body    => Fog::JSON.encode(:application_credential => credential)
+              )
+            end
+          end
+  
+          class Mock
```
geemus, line 18: For consistency with other parts of fog, could you add a method to each of the mock versions that has the same signature, and whose body is:
```
raise Fog::Errors::MockNotImplemented
```

That way it helps ensure that the behavior is obvious/consistent, as opposed to potentially NoMethod errors or even silently failing if the method is defined without a body as below.

lib/fog/openstack/identity/v3/requests/get_application_credentials.rb
```diff
@@ -0,0 +1,24 @@
+module Fog
+    module OpenStack
+      class Identity
+        class V3
+          class Real
+            def get_application_credentials(id, user_id)
+              puts "get app creds id: #{id}, from user #{user_id}"
```
geemus, line 7: Looks like another leftover.
```suggestion
```

geemus (approved):
It would also be good to get some tests for this too, now that I think about it. I'll bring this in for now, but we should probably follow up with tests soon if possible to help keep it working. It's just a bit hard to have good tests without mocks, and that's a whole other chunk of work.

geemus:
Plus the main test suite has been failing for some time now, so also not ideal.

geemus:
@andrejcermak did you need a release with this?

andrejcermak:
Yes, I intend to use this in [other project](https://github.com/OSC/ondemand). I think I can have a look at the tests too.

geemus:
@andrejcermak released in v1.1.5

PR #548, opened 2025-11-28
Minor fixes to compute key_pairs

otuz-godaddy:
Hi - this PR just contains two fixes for some things I caught while using compute keypairs:

* `delete_key_pair()` was accepting an optional parameter `user_id` but never passed it to the API 
* The Mock implementation for `get_key_pair()` was misnamed `delete_key_pair()`

Hopefully it's helpful. Thanks for the library and all the time you put in!


geemus (approved):
Hey, good catches and thanks for fixing these up.

PR #551, opened 2026-08-24
Bump required Ruby version to 2.5

tas50:
The fog core gem already declares `required_ruby_version = '>= 2.5.0'`, so fog-openstack's `'>= 2.2.0'` was never actually satisfiable for anyone installing this gem alongside fog. Ruby 2.2–2.4 are also long past end-of-life.

This aligns fog-openstack with fog core.

### Note on RuboCop

RuboCop infers `TargetRubyVersion` from the gemspec when `.rubocop.yml` doesn't set it explicitly, so this bumps the lint target from 2.2 to 2.5. Verified the offense count is unchanged (2600 before, 2600 after), so no new cops are triggered by the bump.


geemus (approved):
lgtm, thanks!

PR #556, opened 2026-08-28
Migrate the test suite to minitest 6, fix the Volume version fallback, and regenerate the RuboCop baseline

tas50:
Both workflows have been red since CI was added ([run 33117772179](https://github.com/fog/fog-openstack/actions/runs/33117772179)): `rake tests:mock` reported **645 errors**, `rake tests:unit` another 4, and Linting **108 offenses**. This gets everything green on 3.3, 3.4, 4.0 and head.

## Why the tests broke

minitest 6 removed both of the `class Object` monkeypatches that minitest 5 installed, and the suite leaned on each:

- `minitest/spec` no longer does `class Object; include Minitest::Expectations; end`, so a bare `foo.must_equal bar` raises `NoMethodError`. `_(foo).must_equal bar` goes through `Minitest::Expectation` instead.
- `minitest/mock` was extracted into the standalone `minitest-mock` gem, taking `Object#stub` with it.

Every one of the 649 failures was a `NoMethodError` raised *before* the assertion ran — which is also why the run reported only 31 assertions. Nothing was actually being verified.

## Commit 1 — migrate the test suite

- Wrap all **1295** expectation receivers in `_(...)` across `test/`, `spec/` and `unit/`. This is compatible with minitest 5.11+ as well, so it needs no version pin. fog-core's custom `must_match_schema` (registered via `infect_an_assertion`) keeps working unchanged in this form.
- Add the `minitest-mock` development dependency and require it from `unit/test_helper.rb`.
- Replace `must_equal nil` with `must_be_nil`; minitest rejects the former, which the `NoMethodError`s had been hiding.
- Fix a latent bug in `spec/volume_spec.rb`: `must_raise(Excon::Errors::BadRequest, /message/)` was never valid, since minitest splats those arguments straight into a `rescue` clause and a Regexp there is a `TypeError`. That error escaped the `begin`, ran the `ensure` early, and left the `ensure`'s own failure reported in place of the real one.
- Add `rubocop-minitest` and enable `Minitest/GlobalExpectations` so the explicit form does not regress. The rest of the Minitest department stays off, as it is not part of the baseline.

The mechanical rewrite was done with a Prism-based script rather than by hand or by regex, because the receiver has to be wrapped as a whole (`@instance.reload.state.must_equal`, multi-line `proc do ... end.must_raise`). Worth noting that RuboCop's own `Minitest/GlobalExpectations` autocorrect is **not** safe here: it rewrites `proc do ... end.must_raise X` to `_ { proc do ... end }.must_raise X`, which passes the outer block to `must_raise`. Evaluating it just constructs a proc, nothing raises, and the assertion silently stops testing anything.

## Commit 2 — fix the Volume version fallback

`Fog::OpenStack::Volume.new` documents that it returns a V3, V2 or V1 service, and implemented that as:

```ruby
Fog::OpenStack::Volume::V3.new(args) \
|| Fog::OpenStack::Volume::V2.new(args) \
|| Fog::OpenStack::Volume::V1.new(args)
```

`Fog::Service.new` never returns nil or false, so the `||` arms were dead code. On a cloud that does not advertise a `volumev3` endpoint, `V3.new` raises `ServiceTypeError` and the fallback meant to handle exactly that case never runs — the caller gets an exception instead of the V2 or V1 service the cloud does offer.

This rescues `ServiceTypeError` and tries each version in turn, re-raising the last error when no volume service is advertised at all. `unit/volume_test.rb` covers the four outcomes with the versioned classes stubbed, so negotiation is exercised without a token endpoint; both fallback cases fail against the previous implementation.

`spec/volume_spec.rb` no longer runs the shared examples against the version-negotiating entry point. Negotiation authenticates once per version it tries, and those cassettes (recorded in 2016, advertising only `"type": "volume"`) contain a single token request each, so they cannot represent it. V1 and V2 stay covered there.

## Commit 3 — regenerate the RuboCop baseline

The 108 Linting offenses were not new code. `Gemfile.lock` is not committed and the `rubocop` development dependency is unconstrained, so every CI run resolves the newest RuboCop; when a release adds cops, offenses appear in code that has not changed and the baseline stops matching. All 108 fall into four cops that postdate the last baseline: `Style/CollectionCompact`, `Style/HashSlice`, `Style/NumericPredicate` and `Style/SafeNavigation`.

Regenerated with the same flags recorded in the file header:

```
rubocop --auto-gen-config --no-exclude-limit --no-auto-gen-timestamp
```

This also drops the obsolete `Lint/UriEscapeUnescape` entry, which stopped being needed when `URI.encode` was replaced in e51026e. `Minitest/GlobalExpectations` is not suppressed, so it still guards the explicit expectation form.

Note this will recur on future RuboCop releases. Constraining the `rubocop` dependency (or committing a lockfile for CI) would make the Linting job reproducible; happy to add that here or separately if you'd like.

## Verification

```
$ bundle exec rake
759 runs, 733 assertions, 0 failures, 0 errors, 7 skips   # tests:mock
101 runs, 601 assertions, 0 failures, 0 errors, 10 skips  # tests:spec
 57 runs,  60 assertions, 0 failures, 0 errors, 0 skips   # tests:unit

$ bundle exec rubocop
1370 files inspected, no offenses detected
```


geemus (approved):
lgtm, thanks for taking this on and getting us back to green

